PR #14, opened 2025-01-14
feat(click): #ML-684 Add button and repeat parameter to click command

programminx-askui:
Hello, 

I've added following featues:
- click different buttons
- repeat count

### DIFF
--- a/src/askui/agent.py
+++ b/src/askui/agent.py
@@ -17,8 +17,7 @@ from .reporting.report import SimpleReportGenerator
 
 
 class InvalidParameterError(Exception):
-    def __init__(self, *args):
-        super().__init__(*args)
+    pass
 
 class VisionAgent:
     def __init__(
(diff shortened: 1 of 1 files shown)


src/askui/agent.py
```diff
@@ -15,6 +16,10 @@
 from .reporting.report import SimpleReportGenerator
 
 
+class InvalidParameterError(Exception):
+    def __init__(self, *args):
+        super().__init__(*args)
```
programminx-askui: Where do we want to place the exceptions?

src/askui/agent.py
```diff
@@ -45,7 +50,9 @@ def _check_askui_controller_enabled(self) -> None:
                 "AskUI Controller is not initialized. Please, set `enable_askui_controller` to `True` when initializing the `VisionAgent`."
             )
 
-    def click(self, instruction: str, model_name: str | None = None) -> None:
+    def click(self, instruction: str, button: Literal['left', 'middle', 'right'] = 'left', repeat: int = 1, model_name: str | None = None) -> None:
```
programminx-askui, line 52: Do we want to have python inline docs?

adi-wan-askui: Yes

src/askui/agent.py
adi-wan-askui: As close as possible to where they are thrown so this is fine from my perspective for now. But feel free to move it to `scr/askui/exceptions.py` as it is a very generic exception.

Think about extending `ValueError` instead of `Exception` or even using https://docs.pydantic.dev/latest/concepts/validation_decorator  as this provides a very declarative and consistent way for data validation separated from the actual logic of the function.

adi-wan-askui: BTW: You can just do (or do you want to prohibit passing kwargs?)

```suggestion
class InvalidParameterError(Exception):
    pass
```

adi-wan-askui (approved):
Nice job :) 

Added some ideas for optimizations in the comments but no blocker for merging it

src/askui/agent.py
programminx-askui: I stick to ExeptionException but does not move ti to exceptions.py

adi-wan-askui: Okay, but would be good if you check out `ValueError` as this is meant for exactly this kind of exception from my perspective. I don't think we need a custom exception for invalid parameters passed.

PR #15, opened 2025-01-16
feat: #ML-701 Add click without instruction

programminx-askui:
Hello, 

I've added a new feature:
- Click without instruction


adi-wan-askui (changes requested):
See https://github.com/askui/vision-agent/pull/16

PR #16, opened 2025-01-16
Ml 702 agent commands explicit wait commands for a time

programminx-askui:
Hello, 

I've implemented an explicit wait for sec

### DIFF
--- a/src/askui/agent.py
+++ b/src/askui/agent.py
@@ -128,7 +128,6 @@ class VisionAgent:
             >>>     agent.wait(5)  # Pauses execution for 5 seconds.
             >>>     agent.wait(0.5)  # Pauses execution for 500 milliseconds.
         """
-        self._check_askui_controller_enabled()
         time.sleep(sec)
 
     def act(self, goal: str) -> None:
--- a/src/askui/agent.py
+++ b/src/askui/agent.py
@@ -103,7 +103,8 @@ class VisionAgent:
         Pauses the execution of the program for the specified number of seconds.
 
         Args:
-            sec (float): The number of seconds to wait. Must be a non-negative integer.
+            sec (float): The number of seconds to wait. Must be greater than 0.
+            
 
         Raises:
             ValueError: If the provided `sec` is negative.
--- a/src/askui/agent.py
+++ b/src/askui/agent.py
@@ -113,7 +113,7 @@ class VisionAgent:
         return response
     
     @validate_call
-    def wait(self, sec: Annotated[float, Field(gt=0, alias="second")]):
+    def wait(self, sec: Annotated[float, Field(gt=0)]):
         """
         Pauses the execution of the program for the specified number of seconds.
 
@@ -126,6 +126,7 @@ class VisionAgent:
         Example:
             >>> with VisionAgent() agent:
             >>>     agent.wait(5)  # Pauses execution for 5 seconds.
+            >>>     agent.wait(0.5)  # Pauses execution for 500 milliseconds.
         """
         self._check_askui_controller_enabled()
         time.sleep(sec)
--- a/src/askui/agent.py
+++ b/src/askui/agent.py
@@ -68,9 +68,16 @@ class VisionAgent:
         if repeat < 1:
             raise InvalidParameterError("InvalidParameterError! The parameter 'repeat' needs to be greater than 0.")
         self._check_askui_controller_enabled()
+        if self.report is not None:
+            msg = f'click'
+            if button is not 'left':
+                msg = f'{button} ' + msg 
+            if repeat is not 1:
+                msg += f' {repeat}x times'
+            if instruction is not None:
+                msg += f' on "{instruction}"'
+            self.report.add_message("User", msg)
         if instruction is not None:
-            if self.report is not None:
-                self.report.add_message("User", f'click: "{instruction}"')
             logger.debug("VisionAgent received instruction to click '%s'", instruction)
             screenshot = self.client.screenshot() # type: ignore
             x, y = self.model_router.click(screenshot, instruction, model_name)
@@ -104,7 +111,6 @@ class VisionAgent:
 
         Args:
             sec (float): The number of seconds to wait. Must be greater than 0.
-            
 
         Raises:
             ValueError: If the provided `sec` is negative.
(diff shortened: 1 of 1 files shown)


src/askui/agent.py
```diff
@@ -65,14 +68,15 @@ def click(self, instruction: str, button: Literal['left', 'middle', 'right'] = '
...
+        if instruction is not None:
+            if self.report is not None:
+                self.report.add_message("User", f'click: "{instruction}"')
+            logger.debug("VisionAgent received instruction to click '%s'", instruction)
+            screenshot = self.client.screenshot() # type: ignore
+            x, y = self.model_router.click(screenshot, instruction, model_name)
+            if self.report is not None:
+                self.report.add_message("ModelRouter", f"click: ({x}, {y})")
+            self.client.mouse(x, y) # type: ignore
         self.client.click(button, repeat) # type: ignore
```
adi-wan-askui: I think we should add another message to the report just for clicking including button and repetitions info and replace "click" in the message before with "move".  Also it would be nice to have logging for every message added to the rporter, too.

src/askui/agent.py
```diff
@@ -92,6 +96,24 @@ def get(self, instruction: str) -> str:
         if self.report is not None:
             self.report.add_message("Agent", response)
         return response
+    
+    @validate_call
+    def wait(self, sec: Annotated[float, Field(gt=0, alias="second")]):
```
adi-wan-askui: Why the alias? I would remove it.

src/askui/agent.py
```diff
@@ -92,6 +96,24 @@ def get(self, instruction: str) -> str:
...
+        """
+        Pauses the execution of the program for the specified number of seconds.
+
+        Args:
+            sec (float): The number of seconds to wait. Must be a non-negative integer.
```
adi-wan-askui: Either we should change `gt` to `gte` or change the docs here so that they match.

src/askui/agent.py
programminx-askui: For input validation? Or how doe the pydantic validation works without an @validate_call-decorator?

src/askui/agent.py
programminx-askui: I fixed the report of the "User" message now (a little bit up). But we have to differentiate what we log here.

1. The ModelRouter message (`self.report.add_message("ModelRouter",...`)should be moved to the model router itself
2. `self.client.mouse` & `self.client.click` are loged inside the Agent OS

But I would exclude this for now.

programminx-askui: image

src/askui/agent.py
adi-wan-askui: Also we should change `float` to `int` to be in line with documentation `Must be a non-negative integer`.

src/askui/agent.py
adi-wan-askui: Not necessary for input validation. An `alias` is meant for aliasing the parameter name / keyword `sec` so that you can `wait(second=10.0)` instead of `wait(sec=10.0)`. Not necessary from my perspective. I would go with

```suggestion
    def wait(self, sec: Annotated[int, Field(gte=0)]):
```

(using `int` instead of `float` and `gte` instead of `gt` as it says in the documentation, respectively)

programminx-askui: I've choosen float over int and gt=0 over gte=1, because you can also wait 500 milliseconds e.g. `agent.wait(0.5)`. I added an example in the comments.

src/askui/agent.py
```diff
@@ -61,13 +64,28 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
 
         Raises:
             InvalidParameterError: If the 'repeat' parameter is less than 1.
+
+        Example:
+            >>> with VisionAgent() as agent:
+            >>>     agent.click()              # Left click on current position
+            >>>     agent.click("Edit")        # Left click on text "Edit"
+            >>>     agent.click("Edit", button=right)  # Left click on text "Edit"
```
```suggestion
            >>>     agent.click("Edit", button="right")  # Right click on text "Edit"
```

src/askui/agent.py
```diff
@@ -61,13 +64,28 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
...
+            >>>     agent.click(repeat=2)      # Double left click on current position
+            >>>     agent.click("Edit", button="middle", repeat=4)   # 4x middle click on text "edit
```
```suggestion
            >>>     agent.click("Edit", button="middle", repeat=4)   # 4x middle click on text "Edit"
```

src/askui/agent.py
```diff
@@ -61,13 +64,28 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
...
         """
         if repeat < 1:
             raise InvalidParameterError("InvalidParameterError! The parameter 'repeat' needs to be greater than 0.")
         self._check_askui_controller_enabled()
+        if self.report is not None:
+            msg = f'click'
+            if button is not 'left':
+                msg = f'{button} ' + msg 
+            if repeat is not 1:
+                msg += f' {repeat}x times'
```
adi-wan-askui: msg = 'click'
            if button is not 'left':
                msg = f'{button} {msg}' 
            if repeat != 1:
                msg += f' {repeat}x times'

src/askui/agent.py
```diff
@@ -93,6 +111,25 @@ def get(self, instruction: str) -> str:
...
+
+        Raises:
+            ValueError: If the provided `sec` is negative.
+
+        Example:
+            >>> with VisionAgent() agent:
+            >>>     agent.wait(5)  # Pauses execution for 5 seconds.
+            >>>     agent.wait(0.5)  # Pauses execution for 500 milliseconds.
+        """
+        self._check_askui_controller_enabled()
```
adi-wan-askui: Why the check? Would remove it so that there is no dependency on controller.

adi-wan-askui (changes requested):

src/askui/agent.py
adi-wan-askui: Works

src/askui/agent.py
adi-wan-askui: Most important is the `repeat != 1` as otherwise a syntax warning is thrown as we are comparing against a literal.

adi-wan-askui (approved):

PR #20, opened 2025-01-22
#ML-705  agent commands keyboard key up down

programminx-askui:
Hello, 

I've implemented:
- `agent.key_up` for release a key
- `agent.key_down` for press a key

### DIFF
--- a/src/askui/agent.py
+++ b/src/askui/agent.py
@@ -131,6 +131,36 @@ class VisionAgent:
         self._check_askui_controller_enabled()
         time.sleep(sec)
 
+    def key_up(self, key: PC_AND_MODIFIER_KEY):
+        """
+        Simulates the release of a key.
+
+        Args:
+            key (PC_AND_MODIFIER_KEY): The key to be released. This can be any key or a combination of keys with modifiers.
+
+        Example:
+            >>> agent.key_up('a')  # Release the 'a' key.
+            >>> agent.key_up('shift')  # Release the 'Shift' key.
+        """
+        if self.report is not None:
+            self.report.add_message("User", f'key_down "{key}"')
+        self.client.keyboard_release(key)
+
+    def key_down(self, key: PC_AND_MODIFIER_KEY):
+        """
+        Simulates the pressing of a key.
+
+        Args:
+            key (PC_AND_MODIFIER_KEY): The key to be pressed. This can be any key or a combination of keys with modifiers.
+
+        Example:
+            >>> agent.key_down('a')  # Press the 'a' key.
+            >>> agent.key_down('shift')  # Press the 'Shift' key.
+        """
+        if self.report is not None:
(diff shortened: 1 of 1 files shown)


src/askui/agent.py
```diff
@@ -131,6 +131,36 @@ def wait(self, sec: Annotated[float, Field(gt=0)]):
         self._check_askui_controller_enabled()
         time.sleep(sec)
 
+    def key_up(self, key: PC_AND_MODIFIER_KEY):
+        """
+        Simulates the release of a key.
+
+        Args:
+            key (PC_AND_MODIFIER_KEY): The key to be released. This can be any key or a combination of keys with modifiers.
```
adi-wan-askui: How would you do a combination at the moment? Can you add an example?

src/askui/agent.py
```diff
@@ -131,6 +131,36 @@ def wait(self, sec: Annotated[float, Field(gt=0)]):
...
+
+        Example:
+            >>> agent.key_up('a')  # Release the 'a' key.
+            >>> agent.key_up('shift')  # Release the 'Shift' key.
+        """
+        if self.report is not None:
+            self.report.add_message("User", f'key_down "{key}"')
```
```suggestion
            self.report.add_message("User", f'key_up "{key}"')
```

src/askui/agent.py
```diff
@@ -131,6 +131,36 @@ def wait(self, sec: Annotated[float, Field(gt=0)]):
...
+        self.client.keyboard_release(key)
+
+    def key_down(self, key: PC_AND_MODIFIER_KEY):
+        """
+        Simulates the pressing of a key.
+
+        Args:
+            key (PC_AND_MODIFIER_KEY): The key to be pressed. This can be any key or a combination of keys with modifiers.
```
adi-wan-askui: Same as above, combination?

src/askui/agent.py
programminx-askui: This is only for holding & releasing a key. 

```python
agent.key("control")
agent.key("a")
```

If you want to do a key combination you can use `agent.keyboard("control", "c")`


So you can do "CTRL+Mouse Click" in VS Code.

src/askui/agent.py
programminx-askui: No. it's holding a key

src/askui/agent.py
adi-wan-askui: Can we then change the documentation accordingly?

```suggestion
            key (PC_AND_MODIFIER_KEY): The key to be released.
```

src/askui/agent.py
adi-wan-askui: With the comment I was referring to that we document here that we can use a combination when this method is actually only meant to be used with a single key.

```suggestion
            key (PC_AND_MODIFIER_KEY): The key to be pressed.
```

PR #23, opened 2025-01-30
Add support for Anthropic Model from AWS Bedrock

### LINKED ISSUE #18: Support for Anthropic Model from AWS Bedrock
Hey team, I prefer using Anthropic model from AWS Bedrock, and believe the community will benefit from this too.

The change is fairly easy too. We need to conditionally check if AWS_PROFILE or ANTHORPIC_API_KEY env variable is set, and depending on it, use `anthropic.AnthropicBedrock()` or `anthropic.Anthropic()`. I've a working version locally. I might be able to submit a patch, if the authors feel like it. Let me know what you feel.

Thanks.

subhamX:
Closes #18

The following pull request add support for AWS bedrock provider. Clients can now either set `ANTHROPIC_BEDROCK_PROFILE` or `ANTHROPIC_API_KEY`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ dependencies = [
     "pillow>=11.0.0",
     "pydantic>=2.9.2",
     "anthropic>=0.37.1",
+    "anthropic[bedrock]>=0.43.1",
     "rich>=13.9.4",
     "pyperclip>=1.9.0",
     "gradio-client>=1.4.3",
--- a/src/askui/models/anthropic/claude.py
+++ b/src/askui/models/anthropic/claude.py
@@ -1,21 +1,20 @@
-import os
 import anthropic
 from PIL import Image
 
 from ...logging import logger
 from ...utils import AutomationError
 from ..utils import scale_image_with_padding, scale_coordinates_back, extract_click_coordinates, image_to_base64
-
+from .claude_provider import ClaudeApiProvider
 
 class ClaudeHandler:
     def __init__(self, log_level):
-        self.model_name = "claude-3-5-sonnet-20241022"
-        self.client = anthropic.Anthropic()
+        claude_api_provider = ClaudeApiProvider()
+        self.client = claude_api_provider.get_api_client()
+        self.model_name = claude_api_provider.get_model_name()
+
         self.resolution = (1280, 800)
         self.log_level = log_level
-        self.authenticated = True
-        if os.getenv("ANTHROPIC_API_KEY") is None:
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
     "pillow>=11.0.0",
     "pydantic>=2.9.2",
     "anthropic>=0.37.1",
```
adi-wan-askui, line 13: Rm as the new `anthropic[bedrock]` takes precedence

src/askui/models/anthropic/claude_provider.py
```diff
@@ -0,0 +1,53 @@
...
+            self.api_client = Anthropic()
+            logger.info("ANTHROPIC_API_KEY is set, using Anthropic API provider.")
+        elif anthropic_bedrock_profile:
+            self.api_provider = APIProvider.BEDROCK
+            self.api_client = AnthropicBedrock(
+                aws_profile=anthropic_bedrock_profile,
+            )
+            logger.info("ANTHROPIC_BEDROCK_PROFILE is set, using Anthropic Bedrock API provider.")
+        else:
+            raise ValueError("ANTHROPIC_API_KEY or ANTHROPIC_BEDROCK_PROFILE must be set")
```
adi-wan-askui, line 42: It is fine if they are not set as there might be other models available. So either a debug or info log is fine from my perspective.

src/askui/models/anthropic/claude.py
```diff
@@ -1,21 +1,20 @@
...
-        self.client = anthropic.Anthropic()
+        claude_api_provider = ClaudeApiProvider()
+        self.client = claude_api_provider.get_api_client()
+        self.model_name = claude_api_provider.get_model_name()
+
         self.resolution = (1280, 800)
         self.log_level = log_level
-        self.authenticated = True
-        if os.getenv("ANTHROPIC_API_KEY") is None:
-            self.authenticated = False
```
adi-wan-askui, line 18: `authenticated` property cannot be removed as the `ModelRouter` logic depends on it. Instead would extend it to account for the new env variable to "authenticate" (maybe you can use a property or method exposed by api client to check if it is really authenticated but also fine to just check the env variable but would also encapsulate that in the provider similar to client and model name)

adi-wan-askui (changes requested):
Hey @subhamX, sorry for the time it took to get to you PR.

Thanks for the awesome contribution! Solid work :)

Just some minor remarks before we are able to merge it.

PR #28, opened 2025-02-25
Ml 685 agent commands move mouse to target without click

programminx-askui:
Hello, 

I added a `mouse_move` command


src/askui/agent.py
```diff
@@ -88,13 +88,36 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
...
-            self.client.mouse(x, y) # type: ignore
+            self.__mouse_move(instruction, model_name)
         self.client.click(button, repeat) # type: ignore
 
+    def __mouse_move(self, instruction: str, model_name: Optional[str] = None) -> None:
+        self._check_askui_controller_enabled()
+        screenshot = self.client.screenshot() # type: ignore
+        x, y = self.model_router.click(screenshot, instruction, model_name)
+        if self.report is not None:
+            self.report.add_message("ModelRouter", f"click: ({x}, {y})")
```
adi-wan-askui: Is this on purpose that it still says `click`? Is the reasoning because this is the method of the model router we called which is about predicting a position of an element?


Currently, I think this can be confusing to the user and is also not reflecting clearly what we do. What do you think? What do you think about going with something like `locate` instead of `click` and renaming `instruction` to `locator` to make it clearer to the user what this is and differentiate it from an actual instruction of doing something and from actual clicking?

(Later, I think it would make sense to expose a `locate` method directly and a `mouse_move` that takes `x`, `y` instead of or aside from `instruction` to make the methods more focusing on a single responsibility making it easier for users to build with them automations with the downside of being more verbose / needing to write more code)

src/askui/models/router.py
```diff
@@ -6,21 +7,72 @@
...
+    return response
+
+class GroundingModelRouter(ABC):
+
+    @abstractmethod
+    def click(self, screenshot: Image.Image, instruction: str, model_name: str | None = None):
+        pass
+
+    @abstractmethod
+    def is_responsible(self, model_nam: Optional[str]):
```
adi-wan-askui:
```suggestion
    def is_responsible(self, model_nam: Optional[str]) -> bool:
```

Saw some more type hints missing. Let's add typechecking as soon as possible using mypy so that we don't forget the type hints.

src/askui/models/router.py
```diff
@@ -6,21 +7,72 @@
...
 
+def handle_response(response: tuple[int | None, int | None], instruction: str):
+    if response[0] is None or response[1] is None:
+        raise AutomationError(f'Could not locate "{instruction}"')
+    return response
+
+class GroundingModelRouter(ABC):
+
+    @abstractmethod
+    def click(self, screenshot: Image.Image, instruction: str, model_name: str | None = None):
```
adi-wan-askui:
```suggestion
    def click(self, screenshot: Image.Image, instruction: str, model_name: str | None = None) -> tuple[int, int]:
```

src/askui/models/router.py
```diff
@@ -39,44 +91,28 @@ def get_inference(self, screenshot: Image.Image, instruction: str, model_name: s
...
-            logger.debug(f"Routing click prediction to askui-ocr")
-            x, y = self.askui.click_ocr_prediction(screenshot, instruction)
-            return self.handle_response((x, y), instruction)
-        if self.askui.authenticated and model_name == "askui-combo":
-            logger.debug(f"Routing click prediction to askui-combo")
-            x, y = self.askui.click_pta_prediction(screenshot, instruction)
-            if x is None or y is None:
-                x, y = self.askui.click_ocr_prediction(screenshot, instruction)
-            return self.handle_response((x, y), instruction)
+            return handle_response((x, y), instruction)
```
adi-wan-askui: Like the extraction + using polymorphism instead of the if conditions.

src/askui/models/router.py
adi-wan-askui:
```suggestion
    def is_responsible(self, model_name: Optional[str]) -> bool:
```

src/askui/models/router.py
```diff
@@ -39,44 +91,28 @@ def get_inference(self, screenshot: Image.Image, instruction: str, model_name: s
...
         if self.claude.authenticated and model_name == "anthropic-claude-3-5-sonnet-20241022":
             logger.debug("Routing click prediction to Anthropic")
             x, y = self.claude.click_inference(screenshot, instruction)
-            return self.handle_response((x, y), instruction)
+            return handle_response((x, y), instruction)
+        
+        for grounding_model_router in self.grounding_model_routers:
+            if grounding_model_router.is_responsible(model_name) and grounding_model_router.is_authenticated():
+                return grounding_model_router.click(screenshot, instruction, model_name)
```
adi-wan-askui: Nicely extendable

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,40 +1,108 @@
...
+    """
+    try:
+        with open(file_path, 'r') as file:
+            return json.load(file)
+    except FileNotFoundError:
+        print(f"Error: File '{file_path}' not found")
+        raise
+    except json.JSONDecodeError as e:
+        print(f"Error: Invalid JSON in file '{file_path}': {str(e)}")
+        raise
```
adi-wan-askui: Would remove the print statements as we let the error bubble up anyway or raise new errors with all info included, e.g., original error + file path, or at least replace the prints using proper logging

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,40 +1,108 @@
...
 PC_AND_MODIFIER_KEY = Literal['command', 'alt', 'control', 'shift', 'right_shift', 'backspace', 'delete', 'enter', 'tab', 'escape', 'up', 'down', 'right', 'left', 'home', 'end', 'pageup', 'pagedown', 'f1', 'f2', 'f3', 'f4', 'f5', 'f6', 'f7', 'f8', 'f9', 'f10', 'f11', 'f12', 'space', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '!', '"', '#', '$', '%', '&', "'", '(', ')', '*', '+', ',', '-', '.', '/', ':', ';', '<', '=', '>', '?', '@', '[', '\\', ']', '^', '_', '`', '{', '|', '}', '~']
 
 
+HAS_ASKUI_COMPONENT_REGISTRY = os.getenv("ASKUI_COMPONENT_REGISTRY_FILE") is not None
```
adi-wan-askui: Would read the whole environment variable once, e.g., 

```suggestion
ASKUI_COMPONENT_REGISTRY_FILE = os.getenv("ASKUI_COMPONENT_REGISTRY_FILE")
```

instead of reading it multiple times and just replace the conditions with checks if this is empty (None or empty string)

src/askui/agent.py
```diff
@@ -88,13 +88,36 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
             self.report.add_message("User", msg)
         if instruction is not None:
             logger.debug("VisionAgent received instruction to click '%s'", instruction)
```
adi-wan-askui: Should we not make this similar to mouse moving adding a message to the report á la "User click"?

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,40 +1,108 @@
...
+        
+        
+        askui_remote_device_controller_path = execcutables.get('AskUIRemoteDeviceController')
+        if askui_remote_device_controller_path is None:
+            raise ValueError("AskUIRemoteDeviceController executables does not exists.")
+        
+        if not os.path.isfile(askui_remote_device_controller_path):
+            raise FileNotFoundError(f"AskUIRemoteDeviceController executable does not exits under '{askui_remote_device_controller_path}'")
+        
+        return askui_remote_device_controller_path
```
adi-wan-askui: I propose we use pydantic base settings instead of reading the environment variable in which we can define the structure and validation (at least mostly) declaratively and more concisely.

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,40 +1,108 @@
...
+
+        if component_registry["DefinitionVersion"] != "1":
+            raise ValueError("ValueError! Invalid AskUIComponentRegistry DefinitionVersion format: ", component_registry["DefinitionVersion"])
+        
+
+        installed_packages = component_registry.get('InstalledPackages')
+        if installed_packages is None:
+            raise ValueError("InstalledPackages not found in the component registry.")
+        
+        remote_device_controller_package_id = "{aed1b543-e856-43ad-b1bc-19365d35c33e}"
```
adi-wan-askui: What about moving this to a constant?

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,40 +1,108 @@
...
+        remote_device_controller = installed_packages.get(remote_device_controller_package_id)
+
+        if remote_device_controller is None:
+            raise ValueError("RemoteDeviceController is not installed in the component registry.")
+        
+        execcutables = remote_device_controller.get('Executables')
```
adi-wan-askui:
```suggestion
        executables = remote_device_controller.get('Executables')
```

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,40 +1,108 @@
...
+    def __find_remote_device_controller_by_legacy_path(self) -> str:
         if sys.platform == 'win32':
             return f"{os.environ['ASKUI_INSTALLATION_DIRECTORY']}Binaries\\resources\\assets\\binaries\\AskuiRemoteDeviceController.exe"
         if sys.platform == 'darwin':
             return f"{os.environ['ASKUI_INSTALLATION_DIRECTORY']}/Binaries/askui-ui-controller.app/Contents/Resources/assets/binaries/AskuiRemoteDeviceController"
         return f"{os.environ['ASKUI_INSTALLATION_DIRECTORY']}/Binaries/resources/assets/binaries/AskuiRemoteDeviceController"
-    
+
+
+
```
adi-wan-askui: rm?

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,40 +1,108 @@
...
+        askui_remote_device_controller_path = execcutables.get('AskUIRemoteDeviceController')
+        if askui_remote_device_controller_path is None:
+            raise ValueError("AskUIRemoteDeviceController executables does not exists.")
+        
+        if not os.path.isfile(askui_remote_device_controller_path):
+            raise FileNotFoundError(f"AskUIRemoteDeviceController executable does not exits under '{askui_remote_device_controller_path}'")
+        
+        return askui_remote_device_controller_path
+        
+    def __find_remote_device_controller_by_legacy_path(self) -> str:
```
adi-wan-askui: Just to align us in our coding style: When do you use `__` vs. `_`? I usually go for `_` to keep the class open for extension as I find this hard with the name mangling.

adi-wan-askui (changes requested):
Solid :)

src/askui/models/router.py
programminx-askui: The java way. :-D

src/askui/agent.py
programminx-askui: Let's skip it for now. and refactor it later.

src/askui/tools/askui/askui_controller.py
programminx-askui: `__` it's for me a private method, which should only known within a class. e.g. helper methods which are called from mutliple methods.

src/askui/tools/askui/askui_controller.py
programminx-askui: Gone.

src/askui/tools/askui/askui_controller.py
programminx-askui: Obsolete. Refactored in Pydantic settings

src/askui/tools/askui/askui_controller.py
adi-wan-askui: Okay but why not `_`? Because you want to enforce this by having the name mangling? 

I have bad experience with name mangling (the `__`), e.g., when extending a class and overriding extending methods of the subclass which becomes super ugly (and also unexpected to people coming from other programming languages) which is why I use `_` for private methods, functions etc.

My thoughts kind of align with:  https://stackoverflow.com/a/6930223

What do you think about switching to (single) `_` in the future?

src/askui/agent.py
```diff
@@ -94,7 +94,7 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
     def __mouse_move(self, instruction: str, model_name: Optional[str] = None) -> None:
         self._check_askui_controller_enabled()
         screenshot = self.client.screenshot() # type: ignore
-        x, y = self.model_router.click(screenshot, instruction, model_name)
+        x, y = self.model_router.locate(screenshot, instruction, model_name)
         if self.report is not None:
             self.report.add_message("ModelRouter", f"click: ({x}, {y})")
```
```suggestion
            self.report.add_message("ModelRouter", f"locate: ({x}, {y})")
```

src/askui/agent.py
```diff
@@ -94,7 +94,7 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
     def __mouse_move(self, instruction: str, model_name: Optional[str] = None) -> None:
```
adi-wan-askui, line 94: What about also renaming `instruction` in all of the public api that used `ModelRouter.locate()` underneath to `locator`?

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+import pathlib
+from typing import List, Optional
+from pydantic import UUID4, BaseModel, ConfigDict, Field
+from pydantic.alias_generators import to_camel
+from PIL import Image
+from askui.logging import logger
+
+
+
+class Recangel(BaseModel):
```
adi-wan-askui:
```suggestion
class Rectangle(BaseModel):
```

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    width: int
+    height: int
+
+class AskUIImageMetaData(BaseModel):
+    size: Size
+
+class AiElementJson(BaseModel):    
+    model_config = ConfigDict(
+        alias_generator=lambda field_name: to_camel(field_name),
+        serialization_alias=lambda field_name: to_camel(field_name),
```
adi-wan-askui:
```suggestion
        alias_generator=to_camel,
        serialization_alias=to_camel,
```

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    version: int
+    id: UUID4
+    name: str
+    creation_date_time: datetime
+    image_meta_data: AskUIImageMetaData = Field(alias="image")
+
+class AiElement():
```
adi-wan-askui, line 41: Great place for pydantic BaseModel or `dataclass` (https://docs.python.org/3/library/dataclasses.html) if you don't want the validation for the whole thing ⟶ allows to omit the constructor

(with pydantic BaseModel you may have to set `model_config = ConfigDict(arbitrary_types_allowed=True` because I think it cannot handle `Image.Image` otherwise)

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    rectangle: Recangel
+
+class Size(BaseModel):
+    width: int
+    height: int
+
+class AskUIImageMetaData(BaseModel):
+    size: Size
+
+class AiElementJson(BaseModel):
```
adi-wan-askui: Would rename this to `AiElementMetadata` as there is no json here anymore and to be consistent

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+class Annotation(BaseModel):
+    id: UUID4
+    rectangle: Recangel
+
+class Size(BaseModel):
+    width: int
+    height: int
+
+class AskUIImageMetaData(BaseModel):
```
adi-wan-askui:
```suggestion
class AskUIImageMetadata(BaseModel):
```

for consistency

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    model_config = ConfigDict(
+        alias_generator=lambda field_name: to_camel(field_name),
+        serialization_alias=lambda field_name: to_camel(field_name),
+    )
+
+    version: int
+    id: UUID4
+    name: str
+    creation_date_time: datetime
+    image_meta_data: AskUIImageMetaData = Field(alias="image")
```
adi-wan-askui:
```suggestion
    image_metadata: AskUIImageMetadata = Field(alias="image")
```

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+        
+        addional_ai_element_from_env = []
+        if os.getenv("ASKUI_AI_ELEMENT_LOCATIONS", "") != "":
+            addional_ai_element_from_env = [pathlib.Path(ai_element_loc) for ai_element_loc in os.getenv("ASKUI_AI_ELEMENT_LOCATIONS", "").split(",")],
+        
+        self.ai_element_locations = [
+            pathlib.Path.home() / ".askui" / "SnippingTool" / "AIElement" / workspace_id,
+            *addional_ai_element_from_env,
+            *additional_ai_element_locations
+        ]
```
adi-wan-askui: Great for pydantic base settings ;)

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+        
+        if additional_ai_element_locations is None:
+            additional_ai_element_locations = []
+        
+        addional_ai_element_from_env = []
+        if os.getenv("ASKUI_AI_ELEMENT_LOCATIONS", "") != "":
+            addional_ai_element_from_env = [pathlib.Path(ai_element_loc) for ai_element_loc in os.getenv("ASKUI_AI_ELEMENT_LOCATIONS", "").split(",")],
+        
+        self.ai_element_locations = [
+            pathlib.Path.home() / ".askui" / "SnippingTool" / "AIElement" / workspace_id,
```
adi-wan-askui: Would extract this to a constant, e.g., `DEFAULT_AI_ELEMENTS_PATH`

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+            *addional_ai_element_from_env,
```
adi-wan-askui, line 84:
```suggestion
            *additional_ai_element_from_env,
```

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+            *additional_ai_element_locations
```
adi-wan-askui: Like for the ordering (default > env > param)

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+                    image = Image.open(image_path))
+
+
+class AiElementNotFound(Exception):
+    pass
+
+
+class AiElementCollection:
+
+    def __init__(self, additional_ai_element_locations: Optional[List[pathlib.Path]] = None):
```
adi-wan-askui: Would shorten name to

```suggestion
    def __init__(self, ai_element_locations: Optional[List[pathlib.Path]] = None):
```

I think it is not going to be clear that this adds to env and default without docs anyway and so it is at least consistent with env var name and short

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+        logger.warning("AI Element locations: %s", self.ai_element_locations)
+
+    def find(self, name: str):
+        ai_elements = []
+
+        for location in self.ai_element_locations:
+            path = pathlib.Path(location)
+            
+            json_files = list(path.glob("*.json"))
```
adi-wan-askui: Nice :)

src/askui/models/askui/ai_element_utils.py
```diff
@@ -0,0 +1,108 @@
...
+            
+            if not json_files:
+                logger.warning(f"No JSON files found in: {location}")
+                continue
+                
+            for json_file in json_files:
+                ai_element = AiElement.from_json_file(json_file)
```
adi-wan-askui: Is there currently no convention for the json file naming that allows us to directly select the correct file or an index file?

src/askui/models/askui/api.py
```diff
@@ -32,12 +32,12 @@ def __build_model_composition(self):
...
         return f"{self.inference_endpoint}/api/v3/workspaces/{self.workspace_id}/{endpoint}"
 
-    def predict(self, image: Union[pathlib.Path, Image.Image], instruction: str) -> tuple[int | None, int | None]:
+    def predict(self, image: Union[pathlib.Path, Image.Image], locator: str) -> tuple[int | None, int | None]:
         response = requests.post(
             self.__build_base_url(),
             json={
                 "image": f",{image_to_base64(image)}",
-                **({"instruction": instruction} if instruction is not None else {}),
+                **({"locator": locator} if locator is not None else {}),
```
adi-wan-askui:
```suggestion
                **({"instruction": locator} if locator is not None else {}),
```

or did you rename it in the api, too?

src/askui/models/askui/api.py
```diff
@@ -32,12 +32,12 @@ def __build_model_composition(self):
     def __build_base_url(self, endpoint: str = "inference") -> str:
         return f"{self.inference_endpoint}/api/v3/workspaces/{self.workspace_id}/{endpoint}"
 
-    def predict(self, image: Union[pathlib.Path, Image.Image], instruction: str) -> tuple[int | None, int | None]:
+    def predict(self, image: Union[pathlib.Path, Image.Image], locator: str) -> tuple[int | None, int | None]:
```
adi-wan-askui: like for the consistency :)

src/askui/models/router.py
```diff
@@ -73,46 +73,46 @@ def __init__(self, log_level, report,
...
-    def act(self, controller_client, instruction: str, model_name: str | None = None):
+    def act(self, controller_client, locator: str, model_name: str | None = None):
         if self.tars.authenticated and model_name == "tars":
-            return self.tars.act(controller_client, instruction)
+            return self.tars.act(controller_client, locator)
         if self.claude.authenticated and model_name == "claude":
             agent = ClaudeComputerAgent(controller_client, self.report)
-            agent.run(instruction)
+            agent.run(locator)
         raise AutomationError("Invalid model name for act")
```
adi-wan-askui: I think for `act` it is actually an instruction and not a locator.

src/askui/models/router.py
```diff
@@ -73,46 +73,46 @@ def __init__(self, log_level, report,
...
     
-    def get_inference(self, screenshot: Image.Image, instruction: str, model_name: str | None = None):
+    def get_inference(self, screenshot: Image.Image, locator: str, model_name: str | None = None):
         if self.tars.authenticated and model_name == "tars":
-            return self.tars.get_prediction(screenshot, instruction)
+            return self.tars.get_prediction(screenshot, locator)
         if self.claude.authenticated and model_name == "anthropic-claude-3-5-sonnet-20241022":
-            return self.claude.get_inference(screenshot, instruction)
+            return self.claude.get_inference(screenshot, locator)
         raise AutomationError("Executing get commands requires to authenticate with an Automation Model Provider supporting it.")
```
adi-wan-askui, line 96: I think for the `get_inference` it is actually a `query` (or similar) and not a `locator`

src/askui/tools/askui/askui_controller.py
```diff
@@ -16,78 +17,87 @@
...
+        settings_cls: Type[BaseSettings],
+        init_settings: PydanticBaseSettingsSource,
+        env_settings: PydanticBaseSettingsSource,
+        dotenv_settings: PydanticBaseSettingsSource,
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> Tuple[PydanticBaseSettingsSource, ...]:
+        component_registry_file = os.getenv("ASKUI_COMPONENT_REGISTRY_FILE")
+        if component_registry_file is None:
+             raise AskUISuiteNotInstalledError("AskUI Suite is not installed. Please install it.")
+        return (JsonConfigSettingsSource(settings_cls, component_registry_file),)
```
adi-wan-askui: Awesome that you made the settings works :) 

I have to be honest that I misunderstood the code originally. I thought that `ASKUI_COMPONENT_REGISTRY_FILE` contains the json and not the path to the json. But in this case, I think the following would still be a bit cleaner:

```suggestion
class AskUiComponentRegistry(BaseModel):
    definition_version: int = Field(alias="DefinitionVersion")
    installed_packages: InstalledPackages = Field(alias="InstalledPackages")


class AskUiControllerSettings(BaseSettings):
    model_config = SettingsConfigDict(
        env_prefix="ASKUI_",
    )
    
    component_registry_file: pathlib.Path | None = None
    installation_directory: pathlib.Path | None = None

    @model_validator(mode="after")
    def validate_either_component_registry_or_installation_directory_is_set(self) -> "AskUiControllerSettings":
        if self.component_registry_file is None and self.installation_directory is None:
            raise ValueError("Either ASKUI_COMPONENT_REGISTRY_FILE or ASKUI_INSTALLATION_DIRECTORY environment variable must be set")
        return self

# keys ...

class AskUiControllerServer():
    def __init__(self) -> None:
        self._process = None
        self._settings = AskUiControllerSettings() # type: ignore

    def __find_remote_device_controller(self) -> pathlib.Path:
        if self._settings.installation_directory is not None:
            logger.warning("Outdated AskUI Suite detected. Please update to the latest version.")
            return  self.__find_remote_device_controller_by_legacy_path()
        return self.__find_remote_device_controller_by_component_registry()
    
    def __find_remote_device_controller_by_component_registry(self) -> pathlib.Path:
        assert self._settings.component_registry_file is not None, "Component registry file is not set"
        component_registry = AskUiComponentRegistry.model_validate_json(self._settings.component_registry_file.read_text())
        askui_remote_device_controller_path = component_registry.installed_packages.remote_device_controller_uuid.executables.askui_remote_device_controller
        if not os.path.isfile(askui_remote_device_controller_path):
            raise FileNotFoundError(f"AskUIRemoteDeviceController executable does not exits under '{askui_remote_device_controller_path}'")
        return askui_remote_device_controller_path
        
    def __find_remote_device_controller_by_legacy_path(self) -> pathlib.Path:
        assert self._settings.installation_directory is not None, "Installation directory is not set"
        match sys.platform:
            case 'win32':
                return pathlib.Path(os.path.join(self._settings.installation_directory, "Binaries", "resources", "assets", "binaries", "AskuiRemoteDeviceController.exe"))
            case 'darwin':
                return pathlib.Path(os.path.join(self._settings.installation_directory, "Binaries", "askui-ui-controller.app", "Contents", "Resources", "assets", "binaries", "AskuiRemoteDeviceController"))
            case 'linux':
                return pathlib.Path(os.path.join(self._settings.installation_directory, "Binaries", "resources", "assets", "binaries", "AskuiRemoteDeviceController"))
            case _:
                raise NotImplementedError(f"Platform {sys.platform} not supported by AskUI Remote Device Controller")

    def __start_process(self, path):
        self._process = subprocess.Popen(path)
        wait_for_port(23000)
        
    def start(self, clean_up=False):
        if sys.platform == 'win32' and clean_up and process_exists("AskuiRemoteDeviceController.exe"):
            self.clean_up()
        self.__start_process(str(self.__find_remote_device_controller()))

    def clean_up(self):
        if sys.platform == 'win32':
            subprocess.run("taskkill.exe /IM AskUI*")
            time.sleep(0.1)

    def stop(self, force=False):
        if force:
            self._process.terminate()
            self.clean_up()
            return
        self._process.kill()
```

adi-wan-askui (changes requested):
Nice :) It is coming together

src/askui/tools/askui/askui_controller.py
programminx-askui: Did I done this in the pastalways wrong? [PEP 8](https://peps.python.org/pep-0008/#naming-conventions) 

1. _single_leading_underscore: weak “internal use” indicator. E.g. from M import * does not import objects whose names start with an underscore.
2. __double_leading_underscore: when naming a class attribute, invokes name mangling (inside class FooBar, __boo becomes _FooBar__boo; see below).


And [it says](https://peps.python.org/pep-0008/#method-names-and-instance-variables): "Use one leading underscore only for non-public methods and instance variables."

So __ is only used for class attributes and _ is only used for instance method and variables. 

I think you're right

programminx-askui: I adapt the code.

src/askui/agent.py
programminx-askui: I thought about it, too.

but I wouldn't do this. I want to avoid that the users have tu deal with the concept of the locator. 

We (PM and me) want to expose as less internals alá GroundingModel and the locator concept, ..., to the users. So they have not to learn new concepts.

src/askui/models/askui/ai_element_utils.py
programminx-askui: I failed to implement this with an BaseModel. Because I need to load the json path. and from then build the image_path and then load the image and the metadata. 

I looked into the [`@compute_filed`](https://docs.pydantic.dev/2.3/usage/computed_fields/)-decorator but I'm not 100 % sure if this is the right. so I desided to go with a class method.

programminx-askui: If you have a better suggestion feel free to change it.

src/askui/models/askui/ai_element_utils.py
programminx-askui: I will move this comment to https://github.com/askui/vision-agent/pull/29

src/askui/models/askui/ai_element_utils.py
programminx-askui: No. Because the path is dependent on the ENV Variable.

programminx-askui: https://github.com/askui/vision-agent/pull/29

src/askui/models/askui/ai_element_utils.py
programminx-askui: The structure is
```
./.askui/SnippingTool/AIElement/<workspace_id>/
  /<uuid4>.json
  /<uuid4>.png
  /<uuid4>.json
  /<uuid4>.png
  /<uuid4>.json
  /<uuid4>.png
```

And you can have multiple AI Elements  for one name. e.g. "jira icon" in 4 different AI Elements

src/askui/models/askui/api.py
programminx-askui: no. I didn't

src/askui/models/router.py
programminx-askui: it's a goal

src/askui/agent.py
adi-wan-askui: I agree that we don't want the user to learn new concepts if we can omit it. But in this case I think we actually have different concepts but just gave them the same name which I think is even less optimal.

- While `act()` actually accepts an `instruction` (either as high level `goal`, e.g., _Book a flight from Frankfurt to San Francisco_, or step by step description of what to do which works better in my experience than just a `goal`, e.g., _Open skyscanner.com, search for flights from Frankfurt to San Francisco, choose the first one etc._), 
- `get()` (or `ask()`) actually needs a query, e.g., _Is there a flight cheaper than $ 1000?_. 
- Last but not least, `click()`, `mouse_move()` etc. are actually about describing the element that you want to interact with, e.g., click on the __magnifying glass icon__. That is why I would go for different names for the parameters:

So I would propose going for the following

- `act()` ⟶ `instruction` (like it better than `goal` as it makes no assumption about how abstract it needs to be)
- `ask()` ⟶ `query`
- `click()`, `mouse_move()` etc. ⟶ `locator` (which for me is a description of the element that you would like to interact with which may also be an image)

Also the term locator is not super new to users that are engineers working on automation (RPA, test etc.), see
- https://playwright.dev/docs/api/class-locator
- https://www.selenium.dev/documentation/webdriver/elements/locators/
- https://codecept.io/locators/

src/askui/models/askui/ai_element_utils.py
adi-wan-askui: Ahh, true :)

src/askui/models/askui/ai_element_utils.py
adi-wan-askui: I see. We have to keep this in mind for users scaling as this seems not be super performant (parsing each an every ai element json file each time an ai element is used) as soon as someone has 1,000s of AI elements. For now fine.

src/askui/tools/askui/askui_controller.py
adi-wan-askui: Nice, thanks :) 

Yeah, for class methods etc. the name mangling makes sense as you don't want to override (or shadow etc.) them with a subclass but really bind them to the class. Especially since you can call classmethods with `self`

PR #34, opened 2025-04-02
Setup telemetry

adi-wan-askui:


pyproject.toml
```diff
@@ -20,15 +19,22 @@ dependencies = [
...
     "python-dateutil>=2.9.0.post0",
     "openai>=1.61.1",
+    "segment-analytics-python>=2.3.4",
+    "py-machineid>=0.7.0",
+    "httpx>=0.28.1",
 ]
 requires-python = ">=3.10"
 readme = "README.md"
 license = {text = "MIT"}
+dynamic = ["version"]
```
programminx-askui, line 29: What is this?

src/askui/__init__.py
```diff
@@ -1,3 +1,7 @@
+"""AskUI Vision Agent"""
+
+__version__ = "0.2.4"
```
programminx-askui, line 3: I assume the dynamic version

src/askui/agent.py
```diff
@@ -340,16 +353,20 @@ def cli(self, command: str) -> None:
...
         subprocess.run(command.split(" "))
 
-    def close(self):
+    @telemetry.record_call()
+    def close(self) -> None:
         if self.client:
             self.client.disconnect()
         if self.controller:
             self.controller.stop(True)
+        telemetry.flush()
```
programminx-askui: What's happend when one of the methods above throws an exception?

src/askui/telemetry/pkg_version.py
```diff
@@ -0,0 +1,14 @@
+from importlib.metadata import PackageNotFoundError, version
+
+from askui.logger import logger
+
+
+PKG_NAME = "askui"
```
programminx-askui: Why do we need to define the package name here? Can we not retrive this dynamically?

src/askui/telemetry/telemetry.py
```diff
@@ -0,0 +1,229 @@
...
+    app_version: str = get_pkg_version()
+    group_id: str | None = Field(
+        default=os.environ.get("ASKUI_WORKSPACE_ID"),
+        description=(
+            'The group ID of the user. Defaults to the "ASKUI_WORKSPACE_ID" environment variable if set, '
+            "otherwise `None`."
+        ),
+    )
+    device_id: str = Field(
+        default_factory=lambda: map_guid_to_uuid4(machineid.hashed_id("askui")),
```
programminx-askui: Are you using the same as @mlikasam-askui? In some cases the OS function is deactivated. So the machineid can throw an exception. How do you handle this?

src/askui/telemetry/processors.py
```diff
@@ -0,0 +1,110 @@
...
+    context: TelemetryContext
+    timestamp: datetime
+
+
+class SegmentSettings(BaseModel):
+    api_url: HttpUrl = HttpUrl("https://tracking.askui.com/v1")
+    write_key: str = "Iae4oWbOo509Acu5ZeEb2ihqSpemjnhY"
+
+
+class Segment(TelemetryProcessor):
```
programminx-askui, line 36: Does Segment support proxy?

src/askui/telemetry/utils.py
```diff
@@ -0,0 +1,41 @@
+import uuid
+
+
+def map_guid_to_uuid4(guid: str | bytes) -> str:
```
programminx-askui: Do we realy need this function? Maybe I have to less knowledge about the specifications. I thought you can pass a GUID directly in the uuid.UUID4 function. 

But maybe I'm wrong.

tests/telemetry/test_telemetry.py
```diff
@@ -0,0 +1,568 @@
+import pytest
+
+from askui.telemetry import InMemoryProcessor, Telemetry, TelemetrySettings
+
+
+def test_telemetry_disabled():
```
programminx-askui: If you adding now tests. Do we want to setup a github actions for testing?

tests/test_version.py
```diff
@@ -0,0 +1,8 @@
+from importlib.metadata import version
+import askui
+
+def test_version_consistency():
+    package_version = version("askui")
+    module_version = askui.__version__
+    assert package_version == module_version, \
```
programminx-askui: Nice.

pyproject.toml
adi-wan-askui: It defines that we use the version from the build backend which is hatch in our case which in turn is configured in its corresponding section to look at `src/askui/__init__.py`. 

Why this change?

We want to follow the convention that users (and also we) can easily check our version inside the code via `askui.__version__` which is common among a lot of python packages. We want this version to be in sync with the version reported by the build backend, e.g., on Pypi, i.e., have a single source of truth instead of maintaining 2 places where the version is defined.

src/askui/agent.py
adi-wan-askui: ...then, this won't be executed. Good catch :) 

Should be fixed now

src/askui/telemetry/pkg_version.py
adi-wan-askui: I decided to do it this was as it was easier and I assumed that it changes rarely if ever. Nevertheless, I think this could be a source of issues if we forget so I switched to dynamically determining this.

src/askui/telemetry/processors.py
adi-wan-askui: Did not test it but assume this to be true as it uses `requests` under the hood.

src/askui/telemetry/telemetry.py
adi-wan-askui: Already discussed with @mlikasam-askui. Currently, not using the same. Chose a more privacy friendly approach here and moved making it consistent across components to https://askyourui.atlassian.net/browse/CL-1317 

Good catch with regards to that machineid can raise an exception. Came up with fallback. I have another talk with @mlikasam-askui about this as I touch `~/.askui`

src/askui/telemetry/utils.py
adi-wan-askui: I use it to derive an anonymous user id from the hashed device guid (that is why the input is actually no guid and the name of the function misleading, updated it). I changed the implementation after discussion with @mlikasam-askui to only use this function for deriving the anonymous id but not the device id to keep the retrieving of device across our components simple and consistent.

tests/telemetry/test_telemetry.py
adi-wan-askui: Wanted to add this later but I added it now as the earlier the better

PR #36, opened 2025-04-03
Add locators

adi-wan-askui:
This PR got a little bigger than I wanted it to get.

The README and tests should allow for good entrypoints to understand the changes.

# Summary

The PR mainly provides support for all kinds of locators that were originally supported by AskUI (see [https://docs-ts.askui.com/docs/api/API/table-of-contents](https://docs-ts.askui.com/docs/api/API/table-of-contents)). Instead of a fluent API, it provides a builder pattern for building locators, e.g., 

instead of 

```typescript
askui.click().textfield().rightOf().text("Password").exec()
```

```python
agent.click(loc.Class("textfield").right_of(loc.Text("Password"))
```

Some naming was changed. Here a list of the most drastic renamings:
- `pta()` ⟶ `Description()`
- `customElement()` ⟶ `Image()`

For details, pls. take a look at the e2e and unit tests which contain tests of all kinds of locators including all kinds of relations, ai elements etc.

Other than that, a lot of the code was refactored to allow for better testing, e.g., allowing for dependency injection to allow for injection of mocks, and less `if-else-conditions`, e.g., using polymorphism where appropriate.

# Open points

- Detailed documentation under [docs.askui.com](docs.askui.com) still needs to be generated and linked within the README

# Interesting directions for the future
- As the model that can be used depends on the locator and vice versa, we should discuss in the future how to properly integrate them. For now, non-string locators are only fully supported by a newly introduced `"askui"` model which chooses an appropriate `"askui"` model. Other models may not at all or only partially support locators. The locators where set up in a way that we can easily extend support for other models in the future.
- The new locators provide the ability to specify the location of an element in relation to multiple other elements, e.g.,
```python
loc.Class("textfield").right_of(loc.Text("Password")).below_of(loc.Text("Sign in"))
```
Currently, this is not supported by most models including `"askui"` but could be an interesting direction in the future. This also allows defining predences using `and_()` and `or_()`, e.g.,
```python
loc.Class("textfield").right_of(loc.Text("Password").or_(loc.Text("Secret"))).and_(loc.Description("Password textfield"))
```


tests/e2e/agent/test_locate_with_relations.py
```diff
@@ -0,0 +1,366 @@
...
+        assert 250 <= y <= 310
+
+    def test_locate_with_description_and_relation(
+        self,
+        vision_agent: VisionAgent,
+        github_login_screenshot: Image.Image,
+        model_name: str,
+    ) -> None:
+        """Test locating elements using description with relation."""
+        locator = Description("Sign in button").below_of(Description("Password field"))
```
adi-wan-askui: Works fine

src/askui/agent.py
```diff
@@ -92,29 +98,34 @@ def click(self, instruction: Optional[str] = None, button: Literal['left', 'midd
...
+            logger.debug("VisionAgent received instruction to click '%s'", locator)
+            self._mouse_move(locator, model_name)
         self.client.click(button, repeat) # type: ignore
-
-    def __mouse_move(self, instruction: str, model_name: Optional[str] = None) -> None:
-        self._check_askui_controller_enabled()
-        screenshot = self.client.screenshot() # type: ignore
-        x, y = self.model_router.locate(screenshot, instruction, model_name)
+        
+    def locate(self, locator: str | Locator, screenshot: Optional[Image.Image] = None, model_name: Optional[str] = None) -> Point:
```
adi-wan-askui, line 109: Locate

PR #39, opened 2025-04-15
Add get cursor position to computer tools

mlikasam-askui:


src/askui/tools/anthropic/computer.py
```diff
@@ -180,8 +180,13 @@ def __call__(
...
             elif action == "cursor_position":
-                # TODO: Implement in the future
-                return ToolError("cursor_position is not implemented by this agent")
+                if self.real_screen_height is None or self.real_screen_width is None:
+                    screenshot = self.controller_client.screenshot(report=False)
+                    self.real_screen_width = screenshot.width
+                    self.real_screen_height = screenshot.height
+                mouse_x, mouse_y = self.controller_client.get_cursor_position()
+                scaled_x, scaled_y = scale_coordinates_forward(mouse_x, mouse_y, self.real_screen_width, self.real_screen_height, self.width, self.height)
+                return ToolResult(output=f"Cursor position: x={scaled_x}, y={scaled_y}")
```
programminx-askui, line 189: FYI: 

All translation of coordination system are handeled in the future inside the agent os.

src/askui/tools/anthropic/computer.py
```diff
@@ -180,8 +180,13 @@ def __call__(
             if action == "screenshot":
                 return self.screenshot()
             elif action == "cursor_position":
-                # TODO: Implement in the future
-                return ToolError("cursor_position is not implemented by this agent")
+                if self.real_screen_height is None or self.real_screen_width is None:
```
adi-wan-askui, line 183: Shouldn't we also check if the display (index) changed?

adi-wan-askui (changes requested):
Good work :) 

Can we rebase this on the current `main` branch and move the `scale_coordinates_forward` (would just rename that to `scale_coordinates`) to https://github.com/askui/vision-agent/blob/e4bbf11a375926a4bddde0e9f444c7c3a54d7236/src/askui/utils/image_utils.py and extract the things that it has in common with `scale_coordinates_back` (determining scale_factor + padding) to a shared function?

programminx-askui (changes requested):

PR #41, opened 2025-04-16
Model Composition / Selection

adi-wan-askui:
- enables users to set model for complete `VisionAgent` context (see README)
- enables users to use `ModelComposition` for askui (as this is very special, I did not include this in the README but would instead write detailed docs in mintlify)

BREAKING CHANGES:
- model value "claude" for `VisionAgent.act()` changed to
  "anthropic-claude-3-5-sonnet-20241022" to be consistent with rest of possible model values and README
- `model_name` parameter renamed to `model`


tests/e2e/agent/test_model_composition.py
```diff
@@ -0,0 +1,111 @@
...
+        model: ModelComposition,
+    ) -> None:
+        if any("trained" not in m.tags for m in model):
+            pytest.skip("Skipping test for non-trained model")
+        with pytest.raises(Exception):
+            vision_agent.locate(Text("Sign in", text_type="exact"), github_login_screenshot, model=model)
+            vision_agent.locate(Text("Sign in", text_type="regex"), github_login_screenshot, model=model)
+            vision_agent.locate(Text("Sign in", text_type="contains"), github_login_screenshot, model=model)
+            assert DEFAULT_SIMILARITY_THRESHOLD != 80
+            vision_agent.locate(Text("Sign in", similarity_threshold=80), github_login_screenshot, model=model)
```
adi-wan-askui, line 111: I don't understand this behavior from the askui api, yet. @danyalxahid-askui Do you have more insight? Is this expected?

danyalxahid-askui: @adi-wan-askui The similarity threshold, which we have by default, is at 70. If we want to filter out elements that have a very good confidence in the element detected we adjust the similarity_threshold.

A case could be that there are more "Sign in" detections on the screen, but we only want the ones with a good model confidence we filter them out. This would help more in a stable relation e.g. text right of text "Sign in".

adi-wan-askui: That was not what I was referring to 😆 Sorry for having been unspecific... 

What I was referring to is, why do `exact`, `regex` etc. work with `"trained"` but not untrained model...

PR #54, opened 2025-05-22
Retry on ElementNotFoundError

onur-askui:


src/askui/agent.py
```diff
@@ -56,6 +65,7 @@ def __init__(
         reporters: list[Reporter] | None = None,
         tools: AgentToolbox | None = None,
         model: ModelComposition | str | None = None,
+        retry: Retry = DEFAULT_RETRY,
```
adi-wan-askui: Even though declared as a constant this is discouraged in Python to pass mutable things, e.g., instances of a class, as default arguments as someone could mutate it with the sideeffect of every `VisionAgent` instance being affected. Instead I would just allow passing `None` and init a fresh instance inside the constructor.
```suggestion
        retry: Retry | None = None,
```

src/askui/agent.py
```diff
@@ -73,6 +83,7 @@ def __init__(
             else model_router
         )
         self.model = model
+        self._retry = retry
```
```suggestion
        self._retry = retry or Retry(
            strategy="Exponential",
            base_delay=1000,
            retry_count=3,
            on_exception_types=(ElementNotFoundError,),
        )
```

src/askui/retry.py
```diff
@@ -0,0 +1,85 @@
...
+    stop_after_attempt,
+    wait_exponential,
+    wait_fixed,
+    wait_incrementing,
+)
+
+from askui.logger import logger
+
+
+class Retry:
```
adi-wan-askui: Would define this as an abstract class (`abc.ABC`) with the abstract `attemp()` method and export this through `__init__.py` so that users can import it and implement it however they want. The current implementation becomes then our default implementation of that abstract class. Would also export it so that users can also choose to just init a different instance of `Retry` with different configuration.

src/askui/retry.py
```diff
@@ -0,0 +1,85 @@
...
+    This class provides a flexible way to retry operations that may fail temporarily,
+    supporting different retry strategies (Exponential, Fixed, Linear) and configurable
+    parameters for delay and retry count.
+
+    Args:
+        on_exception_types (Tuple[Type[Exception]]): Tuple of exception types that should trigger a retry
+        strategy (Literal["Exponential", "Fixed", "Linear"]): The retry strategy to use:
+            - "Exponential": Delay increases exponentially between retries
+            - "Fixed": Constant delay between retries
+            - "Linear": Delay increases linearly between retries
```
adi-wan-askui:
```suggestion
            - `"Exponential"`: Delay increases exponentially between retries
            - `"Fixed"`: Constant delay between retries
            - `"Linear"`: Delay increases linearly between retries
```

Just so that it is nicely formatted in our docs :)

src/askui/retry.py
```diff
@@ -0,0 +1,85 @@
...
+            return wait_incrementing(self._base_delay / 1000)
+        return wait_exponential(multiplier=self._base_delay / 1000)
+
+    def my_before_sleep(self, retry_state: RetryCallState) -> None:
+        logger.info(
+            "Retrying %s: attempt %s ended with: %s",
+            retry_state.fn,
+            retry_state.attempt_number,
+            retry_state.outcome,
+        )
```
adi-wan-askui: Like that you thought of logging and how generically you did it :)

src/askui/retry.py
```diff
@@ -0,0 +1,85 @@
...
+        self,
+    ) -> wait_fixed | wait_incrementing | wait_exponential:
+        """Get the appropriate wait strategy based on the configured retry strategy."""
+        if self._strategy == "Fixed":
+            return wait_fixed(self._base_delay / 1000)
+        if self._strategy == "Linear":
+            return wait_incrementing(self._base_delay / 1000)
+        return wait_exponential(multiplier=self._base_delay / 1000)
+
+    def my_before_sleep(self, retry_state: RetryCallState) -> None:
```
adi-wan-askui: Curious: What was the reasoning behind method name?

src/askui/retry.py
```diff
@@ -0,0 +1,85 @@
...
+    Example:
+        ```python
+        retry = Retry(
+            on_exception_types=(ConnectionError, TimeoutError),
+            strategy="Exponential",
+            base_delay=1000,
+            retry_count=3
+        )
+        result = retry.attempt(some_function)
+        ```
```
adi-wan-askui: Awesome that you included examples

src/askui/retry.py
adi-wan-askui: My not be formatted well by our docu generator. We have to check when we generate it.

src/askui/retry.py
onur-askui: Makes totally sense

src/askui/retry.py
onur-askui: How can I test that?

onur-askui: How can I test that?

src/askui/retry.py
onur-askui: That's just a terrible name, leftover from the experiments with the lifecycle hooks of the lib. Good call.

src/askui/retry.py
adi-wan-askui: https://github.com/askui/docs-mintlify?tab=readme-ov-file#update-pages

src/askui/retry.py
```diff
@@ -0,0 +1,121 @@
...
+    """  # noqa: E501
+
+    def __init__(
```
adi-wan-askui, line 84: Is the `validate_call` decorator missing?

src/askui/retry.py
```diff
@@ -0,0 +1,121 @@
...
+        Returns:
+            The result of the function execution
+
+        Raises:
+            Exception: Any exception that occurs during execution after
+                      all retry attempts are exhausted
+        """
+
+
+class RetryPolicy(Retry):
```
adi-wan-askui: Not a veto from my side for a merge but not yet 100% happy with the name as `Retry` and `RetryPolicy` feel like kind of the same to me. I would not expect `RetryPolicy` to be an instance of `Retry` just from reading. What was the reasoning behind it? What about `DefaultRetry` or `ConfigurableRetry`?

adi-wan-askui (changes requested):
I like it better :)

src/askui/retry.py
onur-askui: Good thing you mention. I'm also not quite happy with it. For a `DefaultRetry` I think it's too flexible. About `ConfigurableRetry` I thought: well it's rather parameterizing then configuring so went with `Policy`.

We can go with `ConfigurableRetry`.

adi-wan-askui: Totally see your points. I leave it up to you. Would also be fine with how it is for now.

adi-wan-askui (approved):

src/askui/retry.py
onur-askui: Renamed to ConfigurableRetry [:

PR #58, opened 2025-05-27
Resolve issue where OS agent fails to stop on close

mlikasam-askui:


adi-wan-askui (approved):
Awesome, thx :)

src/askui/tools/askui/askui_controller.py
```diff
@@ -200,7 +200,7 @@ def stop(self, force: bool = False) -> None:
             force (bool, optional): Whether to forcefully terminate the process.
                 Defaults to `False`.
         """
-        if not hasattr(self, "process") or self._process is None:
+        if not hasattr(self, "_process") or self._process is None:
```
onur-askui: The `hasattr` check seems to be redundant to me because `_process` is always set in `__init__`. Do you see a reason to keep it?

```suggestion
        if self._process is None:
```

mlikasam-askui: Yep, 100% agree.
Could you refactor your code suggestion to remove empty lines so I can commit it?

PR #64, opened 2025-06-03
refactor/split chat in backend frontend

adi-wan-askui:
- feat(mcp): setup mcp app for vision agent
- refactor(chat): extract threads and messages api
- feat(chat): add thread deletion
- feat(tools)!: add `PynputAgentOs`
- feat(tools,chat): add click event listening/recording
- chore: add cursor rules
- feat(chat): wrap api into fastapi app
- refactor(models): extract methods etc. in computer agents
- feat(agent)!: allow passing messages and callbacks to `VisionAgent.act()`


src/askui/chat/api/messages/service.py
```diff
@@ -0,0 +1,172 @@
...
+
+    id: str = Field(default_factory=lambda: generate_time_ordered_id("msg"))
+    thread_id: str
+    created_at: AwareDatetime = Field(
+        default_factory=lambda: datetime.now(tz=timezone.utc)
+    )
+    object: str = "message"
+
+
+class MessageEvent(Event):
```
onur-askui, line 23: Naming convention idea for events: Past-tense version of the command (verb).

Edit: I see now down in `RunEvent` the intension [:

onur-askui (approved):
Looking great 🚀

src/askui/chat/api/messages/service.py
adi-wan-askui: Totally agree generally. Used resource here as the highest level concept and as all events of a resource are similar for now, went with putting the concrete event into property instead of creating extra classes :)

PR #65, opened 2025-06-04
Fix retry locate to include screenshot

onur-askui:


adi-wan-askui (approved):
LGTM :) Nice tests :) But was it on purpose to put them into e2e tests? Look more like integration tests to me.

PR #67, opened 2025-06-05
Feat integrate openrouter

onur-askui:


src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class OpenRouterGetModel(GetModel):
+    def __init__(self, settings: OpenRouterSettings):
+        self._settings = settings
+
+        _open_router_key = os.getenv("OPEN_ROUTER_API_KEY")
+        if _open_router_key is None:
+            error_msg = "OPEN_ROUTER_API_KEY is not set"
+            raise ValueError(error_msg)
```
adi-wan-askui, line 23: I would add this to the `OpenRouterSettings` as a property and let `OpenRouterSettings` extend `BaseSettings`. Every property not initialised by the consumer through the constructor is automatically initialized from the env. Is more consistent and allows for more powerful, more declarative validation from my perspective and ensures environment variables are consistently prefixed. Also more powerful as users can actually set then all settings through environment or programmatically.

src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
...
+
+        self._client = OpenAI(
+            api_key=_open_router_key,
+            base_url="https://openrouter.ai/api/v1",
```
adi-wan-askui, line 27: Would add this to the `OpenRouterSettings` (with default value) so if they bring out a new version or switch domain, we don't have to do any code changes.

src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
...
+            ],
+            top_p=None,
+            temperature=None,
+            max_tokens=150,
+            stream=False,
+            seed=None,
+            stop=None,
+            frequency_penalty=None,
+            presence_penalty=None,
+        )
```
adi-wan-askui, line 55: Can we create a `BaseModel` (maybe `MessagesCreateSettings` or `ChatCompletionsCreateSettings`) similar to and replacing `ClaudeSettings` we already started? We could add this then as a nested model to `OpenRouterSettings` allowing the user to configure these parameters (except `stream` and `messages`) consistently across different models.

src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
...
+                                "url": image_url,
+                            },
+                        },
+                        {"type": "text", "text": prompt + instruction},
+                    ],
+                }
+            ],
+            top_p=None,
+            temperature=None,
+            max_tokens=150,
```
adi-wan-askui, line 49: Seems low. Would increase this and make it consistent with, e.g., `ClaudeHandler`.

src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
...
+    def get(
+        self,
+        query: str,
+        image: ImageSource,
+        response_schema: Type[ResponseSchema] | None,
+        model_choice: str,
+    ) -> ResponseSchema | str:
+        if response_schema is not None:
+            error_msg = f'Response schema is not supported for model "{model_choice}"'
+            raise NotImplementedError(error_msg)
```
adi-wan-askui, line 68: Couldn't we easily support this (see https://openrouter.ai/docs/features/structured-outputs#best-practices) by setting `require_parameters` to `True` if `response_schema is not None or isinstance(response_schema, str)`? I guess if the model does not support it then, it is going to raise an exception and in all other cases route to a model that supports it.

src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
+import os
```
adi-wan-askui, line 1: I think we should name the file `model.py` or `get_model.py`. This `handler.py` from `anthropic` is a bit misleading and inconsistent with the rest of the naming. WDYT?

src/askui/models/openrouter/prompts.py
```diff
@@ -0,0 +1 @@
+PROMPT_QA = "You are an agent to process screenshots and answer questions about things on the screen or extract information from it. Answer only with the response to the question and keep it short and precise."  # noqa: E501
```
adi-wan-askui, line 1: Can we move this up in the folder structure or into `shared` (coming with the other PR) and reuse it across different models, e.g., in what is currently known as the `ClaudeHandler.get()`?

src/askui/models/openrouter/settings.py
```diff
@@ -0,0 +1,5 @@
+from pydantic import BaseModel, Field
+
+
+class OpenRouterSettings(BaseModel):
+    model: str = Field(..., description="OpenRouter model name")
```
adi-wan-askui, line 5: I would set it per default to `""openrouter/auto"` (see https://openrouter.ai/docs/features/model-routing) and allow everything that is allowed by OpenRouter, including `list[str]` etc. Maybe we can import the type from OpenRouter even... :) WDYT?

src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
...
+from askui.models.models import GetModel
+from askui.models.types.response_schemas import ResponseSchema
+from askui.utils.image_utils import ImageSource
+
+from .prompts import PROMPT_QA
+from .settings import OpenRouterSettings
+
+
+class OpenRouterGetModel(GetModel):
+    def __init__(self, settings: OpenRouterSettings):
```
adi-wan-askui, line 17:
```suggestion
    def __init__(self, settings: OpenRouterSettings | None = None):
```

Would make this optional to make it more easy to use `OpenRouterGetModel` and init the `OpenRouterSettings` internally if `None`.

src/askui/models/openrouter/handler.py
```diff
@@ -0,0 +1,77 @@
...
+from askui.models.exceptions import QueryNoResponseError
+from askui.models.models import GetModel
+from askui.models.types.response_schemas import ResponseSchema
+from askui.utils.image_utils import ImageSource
+
+from .prompts import PROMPT_QA
+from .settings import OpenRouterSettings
+
+
+class OpenRouterGetModel(GetModel):
```
adi-wan-askui, line 16:
```suggestion
class OpenRouterModel(GetModel):
```

Makes it easier for us to extend later

adi-wan-askui (approved):
Super well done (not overengineered, extendable for the most part and well-documented) :) Could not find any blockers from merging, only optimisations I would do for e.g., more consistency or ease of use...

onur-askui:
> Super well done (not overengineered, extendable for the most part and well-documented) :) Could not find any blockers from merging, only optimisations I would do for e.g., more consistency or ease of use... (could also be tackled in follow up PR so we get this out fast)

As discussed, will address the comments in a separate PR

src/askui/models/openrouter/settings.py
onur-askui: Looked into it but couldn't find much. There is https://openrouter.ai/docs/features/provider-routing#json-schema-for-provider-preferences. Experimented with `datamodel-code-generator` to convert jsonschema to models. According to some issues the schema is not super reliable. 

LMK if you have any concrete ideas [:

src/askui/models/openrouter/prompts.py
onur-askui: UI-TARS seems to have different prompts. Should we unify all the prompts across all models? 
https://github.com/askui/vision-agent/blob/main/src/askui/models/ui_tars_ep/prompts.py

adi-wan-askui: Good question. Would say No as the models sometimes require different prompting. Would just see this as a valid default.

src/askui/models/openrouter/settings.py
adi-wan-askui: Nope, fine :)

PR #75, opened 2025-06-13
refactor: OpenRouter model integration and settings

onur-askui:
- Renamed `OpenRouterGetModel` -> `OpenRouterModel`.
- Introduced structured output for OpenRouter.
- Updated the README example to reflect the new model usage.
- Renamed `handler.py` to `model.py` for OpenRouter and Claude.
- Refactored propmpts for Openrouter and Claude.
- Introduced `ChatCompletionsCreateSettings` for managing chat completion settings.
- Moved settings to to `OpenRouterSettings`.
- Added tests for OpenRouter.


src/askui/models/openrouter/model.py
```diff
@@ -0,0 +1,159 @@
...
+        validated_response = _response_schema.model_validate(response_json)
+        return validated_response.root
+
+    @override
+    def get(
+        self,
+        query: str,
+        image: ImageSource,
+        response_schema: Type[ResponseSchema] | None,
+        model_choice: str,
```
onur-askui, line 174: This param has no effect

src/askui/models/shared/settings.py
```diff
@@ -0,0 +1,71 @@
...
+    Args:
+        top_p (float | None, optional): An alternative to sampling with temperature,
+            called nucleus sampling, where the model considers the results of the tokens
+            with top_p probability mass. So 0.1 means only the tokens comprising
+            the top 10% probability mass are considered. We generally recommend
+            altering this or `temperature` but not both.
+
+        temperature (float | None, optional): What sampling temperature to use,
+            between 0 and 2. Higher values like 0.8 will make the output more random,
+            while lower values like 0.2 will make it more focused and deterministic.
```
adi-wan-askui:
```suggestion
        temperature (float | None, optional): What sampling temperature to use,
            between `0` and `2`. Higher values like `0.8` will make the output more random,
            while lower values like `0.2` will make it more focused and deterministic.
```

src/askui/models/shared/settings.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class ChatCompletionsCreateSettings(BaseModel):
+    """
+    Settings for creating chat completions.
+
+    Args:
+        top_p (float | None, optional): An alternative to sampling with temperature,
+            called nucleus sampling, where the model considers the results of the tokens
+            with top_p probability mass. So 0.1 means only the tokens comprising
```
adi-wan-askui:
```suggestion
            with top_p probability mass. So `0.1` means only the tokens comprising
```

src/askui/models/shared/settings.py
```diff
@@ -0,0 +1,71 @@
...
+
+        seed (int | None, optional): If specified, the system will make a best effort
+            to sample deterministically, such that repeated requests with the same seed
+            and parameters should return the same result. Determinism is not guaranteed.
+
+        stop (str | list[str] | None, optional): Up to 4 sequences where the API
+            will stop generating further tokens. The returned text will not contain the
+            stop sequence.
+
+        frequency_penalty (float | None, optional): Number between -2.0 and 2.0.
```
adi-wan-askui:
```suggestion
        frequency_penalty (float | None, optional): Number between `-2.0` and `2.0`.
```

src/askui/models/shared/settings.py
```diff
@@ -0,0 +1,71 @@
...
+            Positive values penalize new tokens based on their existing frequency
+            in the text so far, decreasing the model's likelihood to repeat the same
+            line verbatim.
+
+        presence_penalty (float | None, optional): Number between -2.0 and 2.0.
```
adi-wan-askui:
```suggestion
        presence_penalty (float | None, optional): Number between `-2.0` and `2.0`.
```

src/askui/models/shared/settings.py
```diff
@@ -0,0 +1,71 @@
...
+            with top_p probability mass. So 0.1 means only the tokens comprising
+            the top 10% probability mass are considered. We generally recommend
+            altering this or `temperature` but not both.
+
+        temperature (float | None, optional): What sampling temperature to use,
+            between 0 and 2. Higher values like 0.8 will make the output more random,
+            while lower values like 0.2 will make it more focused and deterministic.
+            We generally recommend altering this or `top_p` but not both.
+
+        max_tokens (int, optional): The maximum number of tokens that can be generated
```
adi-wan-askui, line 22: I think we should tell the user what the optional values default to. Also for the other properties.

src/askui/models/shared/settings.py
```diff
@@ -0,0 +1,71 @@
+from pydantic import BaseModel, Field
+
+
+class ChatCompletionsCreateSettings(BaseModel):
+    """
+    Settings for creating chat completions.
```
adi-wan-askui, line 6: I think in the future it would also be fine to simply include links to where the user can find documentation instead of having to document everything ourselves. But this here has also advantages. WDYT?

src/askui/models/openrouter/model.py
```diff
@@ -0,0 +1,159 @@
...
+                "type": "json_schema",
+                "json_schema": {
+                    "name": "user_json_schema",
+                    "schema": next(
+                        iter(
+                            _response_schema.model_json_schema()
+                            .get("$defs", {})
+                            .values()
+                        )
+                    ),
```
adi-wan-askui: What do we do here? Does that also include the schemas not defined under `"$defs"` as `"$defs"` is only used for nested models as far as I know.

src/askui/models/openrouter/model.py
```diff
@@ -0,0 +1,159 @@
...
+        response_json: object
+        try:
+            response_json = json.loads(str(model_response))
+        except json.JSONDecodeError:
+            error_msg = f"Expected JSON, but model {self._settings.model} returned: {model_response}"  # noqa: E501
+            logger.error(error_msg)
+            raise ValueError(error_msg) from None
+
+        _response_schema = to_response_schema(response_schema)
+        validated_response = _response_schema.model_validate(response_json)
```
adi-wan-askui: Can we optimise this?

adi-wan-askui (changes requested):
Solid work :)

adi-wan-askui (changes requested):
Looking pretty good :)

src/askui/models/openrouter/model.py
onur-askui: As discussed existing patterns adapted.

src/askui/models/openrouter/model.py
onur-askui: `model_validate_json` doesn't seem to be working because we nest data that needs to be validated in `response` which first needs to be decoded.

PR #77, opened 2025-06-18
Refactor tool integration with models

adi-wan-askui:
**BREAKING CHANGES**

- UI Tars model removed from default registry so that users need to register it (custom model) if they want to use it
- Removed `askui.ToolResult`


test.py
```diff
@@ -0,0 +1,107 @@
+import math
```
onur-askui: Scratch file?

adi-wan-askui: Ups, yes :D

PR #80, opened 2025-06-20
Fix android agent

mlikasam-askui:


src/askui/tools/android/agent_os_handler.py
```diff
@@ -7,7 +7,7 @@
 from askui.utils.image_utils import scale_coordinates_back, scale_image_with_padding
 
 
-class PpadbAgentOSHandler(AndroidAgentOs):
+class AndroidAgentOSHandler(AndroidAgentOs):
```
adi-wan-askui:
```suggestion
class AndroidAgentOsHandler(AndroidAgentOs):
```

to be consistent with rest of naming and handling of abbreviations

src/askui/models/shared/tools.py
```diff
@@ -110,6 +110,16 @@ def to_params(
     ) -> list[BetaToolUnionParam]:
         return [tool.to_params() for tool in self._tools]
 
+    def add_tool(self, tool: Tool) -> None:
```
adi-wan-askui:
```suggestion
    def append_tool(self, tool: Tool) -> None:
```

to be clearer on meaning, e.g., if that a tool may override another tool in the map

src/askui/models/shared/tools.py
```diff
@@ -110,6 +110,16 @@ def to_params(
...
+        """Add a tool to the collection."""
+        self._tools.append(tool)
+        self._tool_map[tool.to_params()["name"]] = tool
+
+    def set_tools(self, tools: list[Tool]) -> None:
+        """Set the tools in the collection."""
+        self._tools = tools
+        self._tool_map = {tool.to_params()["name"]: tool for tool in tools}
```
adi-wan-askui:
```suggestion
    def reset_tools(self, tools: list[Tool] | None = None) -> None:
        """Set the tools in the collection."""
        self._tools = tools or []
        self._tool_map = {tool.to_params()["name"]: tool for tool in tools}
```

to be clearer on meaning and more powerful and easy to use for "clearing" case

src/askui/tools/android/agent_os_handler.py
adi-wan-askui: I think we should recall `_os_agent` to `_agent_os` to be clearer what this is.

adi-wan-askui: Makes absolute sense to rename this and as this is not PPadb specific :) 

WDYT about `Facade` instead of `Handler` (including file renaming) to make it even clearer what this is about?

src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -207,7 +207,7 @@ def key_combination(
             error_msg_too_few: str = "Key combination must contain at least 2 keys"
             raise RuntimeError(error_msg_too_few)
 
-        keys_string = " ".join(keys)
+        keys_string = " ".join(key.upper() for key in keys)
```
adi-wan-askui: Is there an advantage of uppercasing the keys here over changing the string literals inside `ANDROID_KEY` to be all uppercased? The second option has the advantage from my perspective of less transformations at runtime and making it more obvious to consumer (agent/user) what key is being used.

src/askui/tools/android/tools.py
```diff
@@ -5,15 +5,15 @@
...
+from askui.tools.android.agent_os_handler import AndroidAgentOSHandler
 
 
 class AndroidScreenshotTool(Tool):
     """
     Takes a screenshot from the currently connected Android device.
     """
 
-    def __init__(self, os_agent_handler: PpadbAgentOSHandler) -> None:
+    def __init__(self, os_agent_handler: AndroidAgentOSHandler) -> None:
```
adi-wan-askui:
```suggestion
    def __init__(self, agent_os: AndroidAgentOs) -> None:
```

WDYT (also for the other tools)? I think it would make this clearer in naming and hide the specific implementation?

src/askui/tools/android/tools.py
```diff
@@ -25,11 +25,11 @@ def __init__(self, os_agent_handler: PpadbAgentOSHandler) -> None:
                 """
             ),
         )
-        self.os_agent_handler = os_agent_handler
+        self._os_agent_handler = os_agent_handler
```
adi-wan-askui:
```suggestion
        self._agent_os = agent_os
```

src/askui/tools/computer.py
```diff
@@ -124,12 +124,10 @@ class ComputerToolBase(Tool, ABC):
     def __init__(
         self,
         agent_os: AgentOs,
-        input_schema: InputSchema,
```
adi-wan-askui: Why remove the input schema?

src/askui/android_agent.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         configure_logging(level=log_level)
         self.os = PpadbAgentOs()
         self._reporter = CompositeReporter(reporters=reporters)
-        android_os_handler = PpadbAgentOSHandler(self.os, self._reporter)
+        android_os_handler = AndroidAgentOSHandler(self.os, self._reporter)
         _models = initialize_default_android_model_registry(
```
adi-wan-askui, line 77: What about assigning the tool collection to `self.tools` or `self.tool_collection` before passing it on to the function? Then, consumers can directly manipulate the tool collection adding and resetting tools and we don't need to add methods to the models or the models facade or the model router or the agent for doing that...

For me the vision for the future (not in this PR) is that we can similar to the `model` parameter pass `tools` to the `act()` (and potentially other method) overriding the default tools set on construction on each `act()` call. WDYT?

src/chat/api/runs/runner/runner.py
```diff
@@ -138,7 +140,21 @@ def _run_human_agent(self, event_queue: queue.Queue[Events]) -> None:
                 )
             )
 
+    def _run_askui_android_agent(self, event_queue: queue.Queue[Events]) -> None:
```
adi-wan-askui, line 143: Well done integrating that :) Going to make that more generic soon but for now this is the way to go.

src/chat/api/assistants/seeds.py
```diff
@@ -11,3 +11,9 @@
...
...AgPHJlY3QgeD0iNTAiIHk9IjMwIiB3aWR0aD0iMTAwIiBoZWlnaHQ9IjYwIiByeD0iMTAiIHJ5PSIxMCIgLz4KCiAgPCEtLSBFeWVzIC0tPgogIDxjaXJjbGUgY3g9IjcwIiBjeT0iNTAiIHI9IjUiIC8+CiAgPGNpcmNsZSBjeD0iMTMwIiBjeT0iNTAiIHI9IjUiIC8+CgogIDwhLS0gQW50ZW5uYXMgLS0+CiAgPGxpbmUgeDE9IjYwIiB5MT0iMzAiIHgyPSI0MCIgeTI9IjEwIiAvPgogIDxsaW5lIHgxPSIxNDAiIHkxPSIzMCIgeDI9IjE2MCIgeTI9IjEwIiAvPgoKICA8IS0tIEJvZHkgLS0+CiAgPHJlY3QgeD0iNDAiIHk9IjkwIiB3aWR0aD0iMTIwIiBoZWlnaHQ9IjEyMCIgcng9IjIwIiByeT0iMjAiIC8+CgogIDwhLS0gIkFza1VJIEFuZHJvaWRBZ2VudCIgbGFiZWwgLS0+CiAgPHRleHQgeD0iMTAwIiB5PSIxNjAiIHRleHQtYW5jaG9yPSJtaWRkbGUiIGZvbnQtc2l6ZT0iMTYiIGZpbGw9ImJsYWNrIiBzdHJva2U9Im5vbmUiIGZvbnQtZmFtaWx5PSJBcmlhbCI+QXNrVUk8L3RleHQ+CiAgPHRleHQgeD0iMTAwIiB5PSIxODAiIHRleHQtYW5jaG9yPSJtaWRkbGUiIGZvbnQtc2l6ZT0iMTIiIGZpbGw9ImJsYWNrIiBzdHJva2U9Im5vbmUiIGZvbnQtZmFtaWx5PSJBcmlhbCI+QW5kcm9pZEFnZW50PC90ZXh0PgoKICA8IS0tIEFybXMgLS0+CiAgPHJlY3QgeD0iMTAiIHk9IjkwIiB3aWR0aD0iMjAiIGhlaWdodD0iMTAwIiByeD0iMTAiIC8+CiAgPHJlY3QgeD0iMTcwIiB5PSI5MCIgd2lkdGg9IjIwIiBoZWlnaHQ9IjEwMCIgcng9IjEwIiAvPgoKICA8IS0tIExlZ3MgLS0+CiAgPHJlY3QgeD0iNjAiIHk9IjIxMCIgd2lkdGg9IjIwIiBoZWlnaHQ9IjYwIiByeD0iNSIgLz4KICA8cmVjdCB4PSIxMjAiIHk9IjIxMCIgd2lkdGg9IjIwIiBoZWlnaHQ9IjYwIiByeD0iNSIgLz4KPC9zdmc+",
```
adi-wan-askui: The text in the svg is a little too much (detail) from my perspective. WDYT about just using an Android icon?

src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -186,7 +187,12 @@ def type(self, text: str) -> None:
             raise RuntimeError(error_msg_nonprintable)
         assert self._selected_display is not None
         display_index: int = self._selected_display.display_index
-        self.shell(f"input -d {display_index} text {text}")
+
+        escaped_text = shlex.quote(text)
+
+        shell_safe_text = escaped_text.replace(" ", "%s")
+
+        self.shell(f"input -d {display_index} text {shell_safe_text}")
```
adi-wan-askui: style: I would remove empty lines

src/askui/tools/android/agent_os.py
```diff
@@ -375,9 +375,9 @@ def set_device_by_index(self, device_index: int = 0) -> None:
         raise NotImplementedError
 
     @abstractmethod
-    def set_device_by_name(self, device_name: str) -> None:
+    def set_device_by_serial_number(self, device_serial_number: str) -> None:
```
adi-wan-askui: As it is already set in the method name (and `sn` is a common abbreviation), i would go for a very short parameter name (also in the method of the `AndroidVisionagent`):
```suggestion
def set_device_by_serial_number(self, dsn: str) -> None:
```

or 

```suggestion
def set_device_by_serial_number(self, device_sn: str) -> None:
```

or 

```suggestion
def set_device_by_serial_number(self, value: str) -> None:
```

adi-wan-askui (approved):
Nice fixes :) 

Well done

src/askui/tools/android/agent_os_handler.py
mlikasam-askui: The AndroidAgentOSHandler was introduced to handle scaling, as the agent cannot operate at the device's full resolution — hence the 'Handler' suffix.
Would 'Facade' be a clearer name than 'Handler'?

src/askui/tools/android/tools.py
mlikasam-askui: It must be an instance of `AndroidAgentOSHandler`; otherwise, using `AndroidAgentOs `directly may lead to inaccuracies due to scaling issues

src/askui/tools/computer.py
mlikasam-askui: it's not used

src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -186,15 +187,18 @@ def type(self, text: str) -> None:
             raise RuntimeError(error_msg_nonprintable)
         assert self._selected_display is not None
         display_index: int = self._selected_display.display_index
-        self.shell(f"input -d {display_index} text {text}")
+
+        escaped_text = shlex.quote(text)
+        shell_safe_text = escaped_text.replace(" ", "%s")
+        self.shell(f"input -d {display_index} text {shell_safe_text}")
```
programminx-askui, line 192: Can we use [shlex ](https://docs.python.org/3/library/shlex.html#shlex.join) here to build the commands?
```suggestion
        self.shell(f"input -d {display_index} text "{shell_safe_text}"")
```

programminx-askui (changes requested):

src/askui/tools/android/ppadb_agent_os.py
mlikasam-askui: Yes, we can use it, but it's not necessary since we control the input in most cases. Also, the quotes aren't needed because spaces are now replaced with `%s`.

src/askui/tools/android/agent_os_handler.py
adi-wan-askui: I think so because it describes what this class does in more detail (wrapping an `AndroidAgentOs` and making it easier to use by handling scaling and other kinds of issues): https://chatgpt.com/share/e/68595322-1704-800f-b4ad-c6c2aa524692

`Handler` is very generic and does not give the same level of meaning (a facade may be a handler but not the other way around)

mlikasam-askui: done

src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -186,15 +187,18 @@ def type(self, text: str) -> None:
             raise RuntimeError(error_msg_nonprintable)
         assert self._selected_display is not None
         display_index: int = self._selected_display.display_index
-        self.shell(f"input -d {display_index} text {text}")
+
```
mlikasam-askui:
```suggestion
```

src/askui/tools/computer.py
adi-wan-askui: WDYT about still keeping it in case we switch to another model which does not have a build-in computer tool and also to make it consistent with rest of tools for now?

FYI: Long term I would rather like to get this through an annotation of the call method or similarly as it can be inferred from the signature and docstring of the method (see https://github.com/jlowin/fastmcp that do it this way). Because currently the writer of the code has to make sure this matches up with the actual method signature which should not be necessary...

src/askui/android_agent.py
```diff
@@ -60,21 +60,22 @@ def __init__(
         configure_logging(level=log_level)
         self.os = PpadbAgentOs()
         self._reporter = CompositeReporter(reporters=reporters)
-        android_os_handler = AndroidAgentOSHandler(self.os, self._reporter)
+        self._act_agent_os_handler = AndroidAgentOsFacade(self.os, self._reporter)
```
adi-wan-askui:
```suggestion
        self._act_agent_os = AndroidAgentOsFacade(self.os, self._reporter)
```

src/askui/android_agent.py
```diff
@@ -60,21 +60,22 @@ def __init__(
...
+        self.act_tool_collection = ToolCollection(
```
adi-wan-askui, line 64:
```suggestion
        self.tools = ToolCollection(
```

Why the act prefix? Is there another tool collection or other tools that we need to differentiate from?

src/askui/tools/android/tools.py
adi-wan-askui: I totally see your point. Still, I think that is something that the tool must not know about but only the initialiser of the tool, i.e., the `AndroidVisionAgent` in this case.

adi-wan-askui (approved):
Nice :)

src/askui/tools/android/tools.py
mlikasam-askui: There was a case where I wanted to override the default tool list while still reusing some of the standard tools and adding additional, more specific ones. By structuring it this way, we're communicating to the user that if they want to use these tools, they must pass an AgentosFacade instance to our standard tools—not the regular agentos instance.

src/askui/android_agent.py
mlikasam-askui: The prefix is intended to clearly communicate to users that these tools are only used by the act command.

Issue #81, opened 2025-06-20
Support open-weight Holo-1 VLM & Runner H UI-agent (high-accuracy, OSS computer-use models)


tristdrum:
### **Motivation / Background**

AskUI Vision Agent already lets users hot-swap between PTA-1, Qwen-VL, Claude Sonnet, etc. (see model matrix in the README).

Since 10 June 2025 the community has an even stronger OSS option:

* **Holo-1** (3 B & 7 B) – best-in-class open-weight UI VLM, ∼76 % screen-element F1.
* **Runner H / Surfer H** – thin REST layer that turns Holo-1 into a fully autonomous computer-use *act* model (LAM), matching Claude’s “Computer Use” API at \~¼ cost.

Adding first-class support will:

* Give users an enterprise-friendly, self-hostable alternative to Anthropic & PTA-1.
* Offer higher accuracy on real-world web UIs than OSS-Atlas / ShowUI.
* Keep Vision Agent the canonical “orchestrator” for every serious UI-automation model out there.

---

### **Proposed Design**

Vision Agent’s architecture already encourages pluggable models via `ActModel`, `GetModel`, `LocateModel`, and the `ModelRegistry` (see “Custom Models” section).
We simply need two thin adapters:

| File                          | Type                                     | Registers as                   | Notes                                                                                                                                   |
| ----------------------------- | ---------------------------------------- | ------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------- |
| `src/askui/models/holo1.py`   | `LocateModel`                            | `"holo-1"`                     | Pure-Python HF adapter<br>uses 🤗 `AutoModelForVision2Seq` + `AutoProcessor`.                                                           |
| `src/askui/models/runnerh.py` | `ActModel` & (`LocateModel` passthrough) | `"runner-h"` (or `"surfer-h"`) | REST client to Runner H SaaS **or** self-hosted container.<br>Mimics Claude Computer-Use JSON schema so upstream logic stays unchanged. |

Both classes are <200 LOC and require **no** change to core agent logic.

---

### **Implementation Steps**

1. **Dependencies**
   *`pyproject.toml`*

   ```toml
   transformers = {extras = ["vision"], version = ">=4.42"}
   huggingface-hub = ">=0.23"
   httpx = {version = ">=0.27", extras = ["http2"]}
   ```
2. **`src/askui/models/holo1.py`**

   ```python
   from askui import LocateModel, Locator, ImageSource, Point
[...]

PR #82, opened 2025-06-23
feat: add Holo-1 Vision Language Model support (partial implementation of #81)

tristdrum:
## Summary
Implements support for Holo-1 Vision Language Model from issue #81. This is a partial implementation as research revealed that Runner H APIs are not publicly accessible.

- ✅ **Holo-1**: Successfully implemented using `Hcompany/Holo1-7B` from Hugging Face
- ❌ **Runner H**: APIs (`api.runnerh.com/v1`) do not exist publicly - appears to be private beta only

### Holo-1 Model (`src/askui/models/huggingface/holo1.py`)
- Uses `AutoModelForImageTextToText` with chat template approach
- Supports GPU (CUDA) and CPU inference with auto-detection
- Parses coordinates from JSON bbox format: `{"bbox": [x1, y1, x2, y2]}`
- Fallback to text coordinate parsing: `"Element at (x, y)"`
- Comprehensive error handling and logging

### Configuration (`src/askui/models/huggingface/settings.py`)
- Environment variables: `HOLO1_MODEL_NAME`, `HOLO1_DEVICE`, etc.
- Default model: `Hcompany/Holo1-7B`
- Auto-detects CUDA availability

### Dependencies Added
- `transformers>=4.45.0`
- `torch>=2.1.0`

### Tests (`tests/integration/models/test_holo1.py`)
- Model initialization and lazy loading
- Element location with string descriptions and Locator objects
- Error handling for missing elements and model loading failures
- Model composition validation

## Research Findings
During implementation, I researched the actual availability of both models:

1. **Holo-1**: Found as `Hcompany/Holo1-7B` on Hugging Face (not `microsoft/holo-1` as initially assumed)
2. **Runner H**: 
   - DNS lookup for `api.runnerh.com` fails
   - Web research shows Runner H is in private beta with waitlist access only
   - No publicly accessible API endpoints

## Test Plan
- [x] Model initializes correctly with proper device detection
- [x] Element location works with mocked transformers inference
- [x] Error handling for invalid models and missing elements
- [x] Code follows project style guidelines (ruff format/lint passes)
- [ ] Integration testing with actual Holo-1 model (requires model download)
- [ ] Performance testing with GPU vs CPU inference

## Next Steps
If Runner H becomes publicly available in the future, its implementation can be added following the same patterns established here.

Closes partially #81

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,8 @@ dependencies = [
     "httpx>=0.28.1",
     "fastmcp>=2.3.4",
     "pure-python-adb>=0.3.0.dev0",
+    "transformers>=4.45.0",
+    "torch>=2.1.0",
 ]
 requires-python = ">=3.10"
 readme = "README.md"
--- a/src/askui/models/huggingface/holo1.py
+++ b/src/askui/models/huggingface/holo1.py
@@ -1,0 +1,224 @@
+"""Holo-1 Vision Language Model implementation for element location.
+
+This module provides the Holo1LocateModel class that uses the Holo-1 VLM
+for locating UI elements on screen based on natural language descriptions.
+"""
+
+import json
+
+from typing_extensions import override
+
+from askui.exceptions import AutomationError, ElementNotFoundError
+from askui.locators.locators import Locator
+from askui.locators.serializers import VlmLocatorSerializer
+from askui.logger import logger
+from askui.models.models import LocateModel, ModelComposition, Point
+from askui.utils.image_utils import ImageSource
+
+
+class Holo1LocateModel(LocateModel):
+    """Holo-1 model implementation for locating UI elements.
+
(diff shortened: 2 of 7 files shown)


onur-askui (changes requested):
Hey @tristdrum,

Thanks for the PR and for bringing up this issue.

It looks like Claude was able to navigate the codebase quite well [:

We need to sort out a couple of things to move forward:
- As you mentioned in the issue, the heavy dependencies need to be optional.
- I'd like to keep the default registry only for models that the askui team offically supports. Currently, we don't really have the capacity to maintain Holo. It would be great if the model could be imported externally.
- We'll also need to add a note about the maintenance as mentioned above.
- We'll need e2e tests.
- And documentation on how to install and get everything running.

Please let me know if I can clarify anything.

PR #84, opened 2025-06-26
feat(agent): add support for locator in `type()`

adi-wan-askui:


src/askui/agent.py
```diff
@@ -294,30 +303,38 @@ def mouse_scroll(
...
+                agent.type("user@example.com", locator="Email")  # Clicks on "Email" input, then types
+                agent.type("password123", locator="Password field", model="custom_model")  # Uses specific model
             ```
         """
-        self._reporter.add_message("User", f'type: "{text}"')
-        logger.debug("VisionAgent received instruction to type '%s'", text)
+        msg = f'type "{text}"'
+        if locator is not None:
+            msg += f" into {locator}"
+            self._click(locator=locator, button="left", repeat=3, model=model)
```
adi-wan-askui: @programminx-askui I think a better way than triple clicking would be to select the whole content using shortcut but for that I think we need to determine the os beforehand as the shortcut is different for windows/linux (ctl+a) and darwin (command+a). Is this fine for now from your perspective?

programminx-askui: Can we have a flag clear=true field? And then check the os? 

I'm not 100% sure that triple click is working on windows.

programminx-askui: It's working on windows. But `clear: bool=true`, would be nice.

PR #92, opened 2025-07-11
feat/credentials fowarding chat

adi-wan-askui:
- feat(models): support setting auth header for AskUI Inference API directly
- feat: allow importing `OnMessageCbParam` type from `askui`
- feat(chat): allow client to configure authorization and workspace id from headers


onur-askui (approved):
🚀 looking great

PR #94, opened 2025-07-11
Update controller grpc

onur-askui:


pyproject.toml
```diff
@@ -56,6 +56,7 @@ typecheck = "mypy"
 "typecheck:all" = "mypy src tests"
 "chat:api" = "uvicorn askui.chat.api.app:app --reload --port 9261"
 "mcp:dev" = "mcp dev src/askui/mcp/__init__.py"
+"grpc:gen" = "bash -c \"python -m grpc_tools.protoc -I src/askui/tools/askui/askui_ui_controller_grpc/proto --python_out=src/askui/tools/askui/askui_ui_controller_grpc/generated --pyi_out=src/askui/tools/askui/askui_ui_controller_grpc/generated --grpc_python_out=src/askui/tools/askui/askui_ui_controller_grpc/generated src/askui/tools/askui/askui_ui_controller_grpc/proto/Controller_V1.proto && sed -i.bak 's/^import Controller_V1_pb2 as Controller__V1__pb2$/from . import Controller_V1_pb2 as Controller__V1__pb2/' src/askui/tools/askui/askui_ui_controller_grpc/generated/Controller_V1_pb2_grpc.py && rm src/askui/tools/askui/askui_ui_controller_grpc/generated/Controller_V1_pb2_grpc.py.bak\""
```
adi-wan-askui: Is see what you meant about readability 😆 

WDYT about moving it an actual bash script, e.g., inside a `<PROJECT_ROOT>/scripts` directory and, then, calling that script from here?

pyproject.toml
```diff
@@ -133,6 +134,7 @@ exclude = [
     "dist",
     "node_modules",
     "venv",
+    "generated"
```
adi-wan-askui: Seems to me like we exclude now every directory and file named `generated`, right? Can we make this more specific by pinning it to that exact path?

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
...
+
+        Returns:
+            controller_v1_pbs.Response_GetDisplayInformation:
+                - displays: List of DisplayInformation objects
+                - virtualScreenRectangle: Overall virtual screen bounds
+
+        Raises:
+            AssertionError: If the gRPC stub is not initialized.
+        """
+        assert isinstance(self._stub, controller_v1.ControllerAPIStub), (
```
adi-wan-askui, line 759: Just note for the future: We should use a `State` pattern to omit this...

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
...
+            "Stub is not initialized"
+        )
+
+        self._reporter.add_message("AgentOS", "get_display_information()")
```
adi-wan-askui, line 763: Note for the future: We should use a decorator that automatically infers this...

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
...
+    ) -> None:
+        """
+        Configure test settings including default capture parameters and delays.
+
+        Args:
+            default_capture_parameters
+                (controller_v1_pbs.CaptureParameters | None, optional):
+                Default capture parameters with displayID and captureArea.
+                Defaults to None.
+            mouse_delay_ms (int, optional): Mouse delay in milliseconds. Defaults to 0.
```
adi-wan-askui: Let's use backticks for code and default parameters in all the docstrings

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
                 )
             ),
         )
+
+    @telemetry.record_call()
+    def get_display_information(
+        self,
+    ) -> controller_v1_pbs.Response_GetDisplayInformation:
```
adi-wan-askui, line 750: Just a comment: I did not expect the methods to be exposed within this PR but nice to have this already. Long-term, I think we should not expose the types from the controller directly in the future (except for experimental apis or short-term workarounds), neither as types of parameters nor reponses, as they as they seem not nicely named, low level etc.

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
...
+    def get_display_information(
+        self,
+    ) -> controller_v1_pbs.Response_GetDisplayInformation:
+        """
+        Get information about all available displays and virtual screen.
+
+        Returns:
+            controller_v1_pbs.Response_GetDisplayInformation:
+                - displays: List of DisplayInformation objects
+                - virtualScreenRectangle: Overall virtual screen bounds
```
adi-wan-askui, line 757: Have never seen that also properties/attributes are documented here. Where did you find this?

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
...
+        response: controller_v1_pbs.Response_GetAutomationTargetList = (
+            self._stub.GetAutomationTargetList(controller_v1_pbs.Request_Void())
+        )
+
+        return response
+
+    @telemetry.record_call()
+    def set_test_configuration(
+        self,
+        default_capture_parameters: controller_v1_pbs.CaptureParameters | None = None,
```
adi-wan-askui: Did you test if this works with `telemetry.record_call()`? I have this question for all methods where we pass more complex parameters here as they all need to be serializable.

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
...
+
+        self._stub.SetActiveAutomationTarget(
+            controller_v1_pbs.Request_SetActiveAutomationTarget(ID=target_id)
+        )
+
+    @telemetry.record_call()
+    def schedule_batched_action(
+        self,
+        action_class_id: controller_v1_pbs.ActionClassID,
+        action_parameters: controller_v1_pbs.ActionParameters,
```
adi-wan-askui, line 936: Here as well: Serializable for telemetry and reporter?

src/askui/tools/askui/askui_controller.py
```diff
@@ -724,3 +725,487 @@ def run_command(self, command: str, timeout_ms: int = 30000) -> None:
...
+            "Stub is not initialized"
+        )
+
+        self._reporter.add_message("AgentOS", "remove_all_actions()")
+
+        self._stub.RemoveAllActions(
+            controller_v1_pbs.Request_RemoveAllActions(sessionInfo=self._session_info)
+        )
+
+    @telemetry.record_call(exclude={"message"})
```
adi-wan-askui: Why exclude it? What can be in that message? What is the use case of the method `send_message()`?

adi-wan-askui (changes requested):
Nice work :) I particularly like
- the tests. Really nice and easy to read and test that every controller method does not fail
- the thorough documentation
- how consistent it is with existent patterns
- that we now have the proto included and clear separation from generated code and proto
- the generation script persisted 

Commented on some things where I would still do something or where I just wanted to share my current thoughts

src/askui/tools/askui/askui_controller.py
onur-askui: Absolutely agree on response types and even some commands. We can sync on the rest in a 1:1 [:

src/askui/tools/askui/askui_controller.py
onur-askui: https://github.com/askui/vision-agent/blob/7f960918edc3a49ad28fea81ed64a9ac424f5a22/src/askui/tools/askui/askui_ui_controller_grpc/proto/Controller_V1.proto#L447

src/askui/tools/askui/askui_controller.py
onur-askui: Agreed. Created CL-1601

src/askui/tools/askui/askui_controller.py
onur-askui: Agreed. Created CL-1602

src/askui/tools/askui/askui_controller.py
onur-askui: `send_message` Is the grpc call to communicate with the new API and message contains the entire payload of a command. 

On second thought, I think that we should remove the telemetry log entirely here. We are already logging on command level. This also gives us more flexibility in terms of what to exclude. see: https://github.com/askui/vision-agent/blob/cc95d38b65cfecd954d5c20551e0d2fd2e851c44/src/askui/tools/askui/askui_controller.py#L1161-L1181

adi-wan-askui: Agree :) Would also rename it to `_send_message` as it seems private.

adi-wan-askui (approved):

PR #95, opened 2025-07-14
fix: override beta flag settings

adi-wan-askui:
- fixes using `AndroidVisionAgent.act()` with Claude 4 behind AskUI API and empty betas (default)
- keep default betas on inference side as `["computer_20241022"]` to be backwards compatible
- allow overriding with empty list
- remove `settings` parameter of `AgentBase.act()` for now from telemetry as it is not fully serializable ⟶ make it serializable in later step

### DIFF
--- a/src/askui/agent_base.py
+++ b/src/askui/agent_base.py
@@ -107,7 +107,7 @@ class AgentBase(ABC):  # noqa: B024
             "locate": model_choice.get("locate", ModelName.ASKUI),
         }
 
-    @telemetry.record_call(exclude={"goal", "on_message"})
+    @telemetry.record_call(exclude={"goal", "on_message", "settings"})
     @validate_call
     def act(
         self,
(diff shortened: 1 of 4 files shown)


src/askui/agent_base.py
```diff
@@ -107,7 +107,7 @@ def _init_model_choice(
             "locate": model_choice.get("locate", ModelName.ASKUI),
         }
 
-    @telemetry.record_call(exclude={"goal", "on_message"})
+    @telemetry.record_call(exclude={"goal", "on_message", "settings"})
```
mlikasam-askui, line 110:
```suggestion
    @telemetry.record_call(exclude={"goal", "on_message", "settings", "tools"})
```

adi-wan-askui: Would like to keep tools (actually also settings but let's move that to the future). Tools should work as they are fully serializable or did you run into issues?

mlikasam-askui: Yes, I ran into issues — it complains because it's a list type

PR #98, opened 2025-07-18
Expose visualisation api

onur-askui:


pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
     "rich>=13.9.4",
     "segment-analytics-python>=2.3.4",
     "tenacity>=9.1.2",
+    "datamodel-code-generator>=0.31.2",
```
adi-wan-askui: Isn't this a dev dependency?

pyproject.toml
```diff
@@ -57,6 +58,7 @@ typecheck = "mypy"
 "chat:api" = "uvicorn askui.chat.api.app:app --reload --port 9261"
 "mcp:dev" = "mcp dev src/askui/mcp/__init__.py"
 "grpc:gen" = "bash -c \"python -m grpc_tools.protoc -I src/askui/tools/askui/askui_ui_controller_grpc/proto --python_out=src/askui/tools/askui/askui_ui_controller_grpc/generated --pyi_out=src/askui/tools/askui/askui_ui_controller_grpc/generated --grpc_python_out=src/askui/tools/askui/askui_ui_controller_grpc/generated src/askui/tools/askui/askui_ui_controller_grpc/proto/Controller_V1.proto && sed -i.bak 's/^import Controller_V1_pb2 as Controller__V1__pb2$/from . import Controller_V1_pb2 as Controller__V1__pb2/' src/askui/tools/askui/askui_ui_controller_grpc/generated/Controller_V1_pb2_grpc.py && rm src/askui/tools/askui/askui_ui_controller_grpc/generated/Controller_V1_pb2_grpc.py.bak\""
+"json:gen" = "datamodel-codegen --output-model-type pydantic_v2.BaseModel --input src/askui/tools/askui/askui_ui_controller_grpc/json_schema/ --input-file-type jsonschema --output src/askui/tools/askui/askui_ui_controller_grpc/generated/"
```
adi-wan-askui, line 59: Same as in other pr. Think I would move it to a bash script

src/askui/tools/askui/askui_controller.py
```diff
@@ -793,9 +789,6 @@ def get_process_list(
         Returns:
             controller_v1_pbs.Response_GetProcessList: Process list response containing:
                 - processes: List of ProcessInfo objects
-
-        Raises:
```
adi-wan-askui, line 797: What was the reason you removed those across all docstrings?

src/askui/tools/askui/command_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+
+def create_set_mouse_position_command(
+    x: int, y: int, session_guid: str
+) -> AskuiAgentosSendRequestSchema:
+    location = Location2(x=Length(root=x), y=Length(root=y))
+    command = Command1(name=Name1.SetMousePosition, parameters=[location])
+
+    header = Header(authentication=Guid(root=session_guid))
+    message = Message(header=header, command=command)
+    return AskuiAgentosSendRequestSchema(message=message)
```
adi-wan-askui, line 93: I think this is a nice place for an `AskuiAgentosSendRequestSchemaBuilder` (class) (builder pattern) that is initialized with the session guid. Could remove a lot of code duplication where each builder method can focus on the actual command. But I think I would just leave it for now as it is :)

src/askui/tools/askui/command_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+
+    return RenderLinePoints(location_points)
+
+
+def create_render_image(base64_data: str) -> RenderImage:
+    return RenderImage(root=base64_data)
+
+
+def create_render_text(text: str) -> RenderText:
+    return RenderText(root=text)
```
adi-wan-askui: What was the reasoning behind wrapping constructors in functions instead of just calling the constructors or model validation directly? (I also asked myself that for other methods than those 2 in the file, e.g., `create_style`)

src/askui/tools/askui/askui_ui_controller_grpc/json_schema/AgentOS-Send-Request-2501.json
```diff
@@ -0,0 +1,354 @@
...
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "authentication": {
+              "$ref": "#/definitions/guid"
+            }
+          },
+          "required": ["authentication"]
+        },
+        "command": {
```
adi-wan-askui, line 130: I think the api is actually pretty nice :) The issue with the bad naming in the code lies more with the generator. WDYT about just going with an LLM (potentially in combination with the generator) for generation? We could easily just create a test that that checks wether the generated pydantic models actually serialize to this json schema `BaseModel.model_json_schema()` and run than generation and test in a loop until test is green (e.g., using agentic IDE).

src/askui/tools/askui/askui_controller.py
onur-askui: First I added them all in the previous PR because we raise an AssertionError so we should document but then changed my mind [: 
IMO this assertion is about our "world view" how the programme should work. Not an exception that should be handled by the consumer. If docstring is seen as a manual of the method for the dev, there is no reason for the AssertionError to be mentioned nor tempt the consumer to handle this case. Can't think of a scenario where this case would be handled programmatically. If the stub is not initialised we should rather crash fast and fix the bug.

It got a little philosophical, for me it's rather nit. Don't mind adding them. Curious about your take on this. [:

src/askui/tools/askui/command_helpers.py
onur-askui: Mix of different things [: 

I think we can/should use `RenderText` `RenderImage` etc wherever they are needed. It should be only in command helpers anyway. The current state is partially a leftover idea from the previous attempt.

When it comes to the `RenderObjectStyle`. I didn't want to expose a code-gen class to the consumer. Plus had the linter yelling at me regarding missing values despite them being all optional. I didn't want to touch the generated code too much.

src/askui/tools/askui/askui_ui_controller_grpc/json_schema/AgentOS-Send-Request-2501.json
onur-askui: That's certainly an idea worth trying. I have 2 not related thoughts on that:
1) Do we want to deal with the client on our end?
2) We might simply not get the same json schema because of how pydantic's code-gen works and deals with nuances.

pyproject.toml
onur-askui: Good point created a dev dependency group and moved datamodel-code-generator, grpcio-tools to it.

src/askui/tools/askui/askui_controller.py
adi-wan-askui: See your point :) 

I lean more to the other side for the reason that a user could actually run into the error if the user did not connect beforehand (called the `connect()` method) and in this case the user just misused the lib and I would like to guide the user to fix that. I think `AssertionError` is just not a very descriptive name of that issue. Should be more something like `ConnectionError`.

But I would be fine with any decision you take.

src/askui/tools/askui/askui_ui_controller_grpc/json_schema/AgentOS-Send-Request-2501.json
adi-wan-askui: 1. Mid- to long-term, no, as Core team is going to provide an SDK, I would leave it for now as it is but it is a nice idea if we have to touch this again until we have the SDK.
2. Totally true although I have very good experience with being able to tune pydantic to match the schema (I have not yet encountered a case where there was no model config option, or field option etc.  I could set to get exactly the schema I want on serialization). Even if this is true, we could instruct the LLM to just disregard certain differences and exclude them from assertions in the test

PR #100, opened 2025-07-24
feat: add ScreenSwitchTool and display information retrieval

danyalxahid-askui:
- Introduced `ScreenSwitchTool` for cycling through multiple displays.
- Added methods in `AgentOs` for retrieving display information and active display.
- Updated `VisionAgent` to utilize the new `ScreenSwitchTool`.
- Enhanced `AskUiControllerClient` with display information retrieval capabilities.


src/askui/agent.py
```diff
@@ -401,6 +402,8 @@ def _get_default_settings_for_act(self, model_choice: str) -> ActSettings:
 
     @override
     def _get_default_tools_for_act(self, model_choice: str) -> list[Tool]:
+        self._tools.append(ScreenSwitchTool(agent_os=self.tools.os))
```
adi-wan-askui, line 405: WDYT about including it right next to the `ExceptionTool` in the constructor? Seems easier and more consistent.

adi-wan-askui, line 405: Aren't we missing the tool for listing all available displays?

src/askui/tools/screen_switch_tool.py
```diff
@@ -0,0 +1,45 @@
+from askui.models.shared.tools import Tool
+from askui.tools.agent_os import AgentOs, DisplayInformation
+
+
+class ScreenSwitchTool(Tool):
```
adi-wan-askui, line 5: Can we

- call the file `display_tools.py` 
- add the tool for listing displays
- rename the tool here to `SetDisplayTool`

so that we are more consistent with our naming across the code base?

src/askui/tools/screen_switch_tool.py
```diff
@@ -0,0 +1,45 @@
...
+            name="screen_switch",
+            description=f"""
+            This tool is useful for switching between multiple displays to find
+            information not present on the current active screen.
+            If more than one display is available, this tool cycles through them.
+            Number of displays available: {len(displays)}.
+            """,
+        )
+        self._agent_os: AgentOs = agent_os
+        self._displays: list[DisplayInformation] = displays
```
adi-wan-askui, line 26: Decent idea of listing the displays here and including it in the tool description although that does not match what is specified in the issue. The problem I see with this is that it is too static as (the number of) displays may change throughout the execution. So I think going with an extra tool for listing displays the better approach. WDYT?

src/askui/tools/screen_switch_tool.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __call__(self) -> None:
```
adi-wan-askui, line 28: I think the tool would be more versatile and easier if it was just forwarding to the `set_display` taking a the  display id as a parameter as then we could just forward without additional logic and we could directly switch to one display without having to call this tool multiple times to switch to the correct display. WDYT?

src/askui/tools/askui/askui_controller.py
```diff
@@ -765,21 +779,22 @@ def get_display_information(
...
+        )
+        return GetDisplayInformationResponse.model_validate(response_dict)
 
     @telemetry.record_call()
     def get_process_list(
         self, get_extended_info: bool = False
     ) -> controller_v1_pbs.Response_GetProcessList:
         """
         Get a list of running processes.
-
```
adi-wan-askui, line 777: Why remove the empty lines across all the docstrings? 😆

src/askui/tools/askui/askui_controller.py
```diff
@@ -723,6 +730,14 @@ def set_display(self, display: int = 1) -> None:
         )
         self._display = display
 
+    @telemetry.record_call()
+    @override
+    def get_active_display(self) -> int:
+        """
+        Get the active display.
+        """
+        return self._display
```
adi-wan-askui, line 739: Well done :) I think we should either include this info in the display information returned by the listing tool / the `GetDisplayInformationResponse` or add an extra tool for this. 

I think it would be great if this was included in the controller api as I think they maintain the source of truth about which display is currently active (see https://askyourui.slack.com/archives/C091TSQ6KP0/p1753431836333729).

src/askui/tools/agent_os.py
```diff
@@ -154,6 +154,26 @@ class ClickEvent(BaseModel):
...
+
+    width: int
+    height: int
+
+
+class DisplayInformation(BaseModel):
+    """Contains information about a single display."""
+
+    display_id: int = Field(alias="displayID")
+    size_in_pixels: SizeInPixels = Field(alias="sizeInPixels")
```
adi-wan-askui:
```suggestion
    display_id: int = Field(validation_alias="displayID")
    size_in_pixels: SizeInPixels = Field(validation_alias="sizeInPixels")
```

as I don't think we need it for serialization.

adi-wan-askui (commented):
Solid for first contribution :)
 
- WDYT about adjusting the system prompt to include a line similar to: `"If you don't see something on one display, list all available displays and see if it can be seen on another screen"`?
- I think we may need to adapt the computer tool as currently it does not adapt the display resolution when another display is selected which may lead to the scaling of coordinates and screenshots not working.

PR #101, opened 2025-07-24
Tars improvements

stephenkopylov:
We’ve fixed a minor issue with the TARS model initialization (it was slightly broken), and added support for the TARS_MODEL_NAME environment variable - previously, it was hardcoded to "tgi".

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,13 +165,13 @@ You can use Vision Agent with UI-TARS if you provide your own UI-TARS API endpoi
 
 **Example Code:**
 ```python
+# Set environment variables before running this code:
+# TARS_URL=http://your-tars-endpoint.com/v1
+# TARS_API_KEY=your-tars-api-key
+# TARS_MODEL_NAME=your-model-name  # Optional, defaults to "ui-tars"
+
 from askui import VisionAgent
 
-# Set environment variables (or set them in your shell)
-import os
-os.environ["TARS_URL"] = "http://your-tars-endpoint.com"
-os.environ["TARS_API_KEY"] = "your-tars-api-key"
-os.environ["TARS_MODEL_NAME"] = "your-model-name"  # Optional, defaults to "ui-tars"
 
 # Use TARS model directly
 with VisionAgent(model="tars") as agent:
--- a/README.md
+++ b/README.md
@@ -161,7 +161,23 @@ You can use Vision Agent with UI-TARS if you provide your own UI-TARS API endpoi
 
 2. Step: Provide the `TARS_URL` and `TARS_API_KEY` environment variables to Vision Agent.
 
-3. Step: Use the `model="tars"` parameter in your `click()`, `get()` and `act()` etc. commands or when initializing the `VisionAgent`.
+3. Step: Use the `model="tars"` parameter in your `click()`, `get()` and `act()` etc. commands or when initializing the `VisionAgent`. The TARS model will be automatically registered if the environment variables are available.
+
+**Example Code:**
+```python
+from askui import VisionAgent
+
+# Set environment variables (or set them in your shell)
+import os
+os.environ["TARS_URL"] = "http://your-tars-endpoint.com"
+os.environ["TARS_API_KEY"] = "your-tars-api-key"
+
+# Use TARS model directly
+with VisionAgent(model="tars") as agent:
+    agent.click("Submit button")  # Uses TARS automatically
+    agent.get("What's on screen?")  # Uses TARS automatically
+    agent.act("Search for flights")  # Uses TARS automatically
+```
 
 
 ## ▶️ Start Building
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ You can use Vision Agent with UI-TARS if you provide your own UI-TARS API endpoi
 
 1. Step: Host the model locally or in the cloud. More information about hosting UI-TARS can be found [here](https://github.com/bytedance/UI-TARS?tab=readme-ov-file#deployment).
 
-2. Step: Provide the `TARS_URL` and `TARS_API_KEY` environment variables to Vision Agent.
+2. Step: Provide the `TARS_URL`, `TARS_API_KEY`, and optionally `TARS_MODEL_NAME` environment variables to Vision Agent.
 
 3. Step: Use the `model="tars"` parameter in your `click()`, `get()` and `act()` etc. commands or when initializing the `VisionAgent`. The TARS model will be automatically registered if the environment variables are available.
 
@@ -171,6 +171,7 @@ from askui import VisionAgent
 import os
 os.environ["TARS_URL"] = "http://your-tars-endpoint.com"
 os.environ["TARS_API_KEY"] = "your-tars-api-key"
+os.environ["TARS_MODEL_NAME"] = "your-model-name"  # Optional, defaults to "ui-tars"
 
 # Use TARS model directly
 with VisionAgent(model="tars") as agent:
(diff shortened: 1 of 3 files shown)


src/askui/models/ui_tars_ep/ui_tars_api.py
```diff
@@ -90,6 +90,11 @@ class UiTarsApiHandlerSettings(BaseSettings):
         min_length=1,
         validation_alias="TARS_API_KEY",
     )
+    tars_model_name: str = Field(
+        default="ui-tars",
```
adi-wan-askui: WDYT about keeping the default `"tgi"` instead of `"ui-tars"` so that we can omit the breaking change?

src/askui/models/model_router.py
```diff
@@ -105,6 +108,36 @@ def hf_spaces_handler() -> HFSpacesHandler:
...
+        logger.debug("TARS model registered successfully")
+
+    except (ValueError, ImportError) as e:
+        # Required TARS environment variables (TARS_URL, TARS_API_KEY) are not set
+        # or TARS module is not available
+        # This is expected behavior, so we just skip TARS registration
+        logger.debug(f"TARS model not registered: {e}")
+
+    return registry
+
```
adi-wan-askui: We like that you tried to include UI Tars in the default model registry so that users can use it right from the get go. We would like to do 2 things differently:

- Remove the dependency on the `tools` package as we are working toward removing all the dependencies of the package `models` on `tools`. If a model requires a certain tool (interface/api), this should be defined in `models` but the actual implementation can then be done in the `tools` package and be injected through method or constructor injection. In this particular case, I think we should go the route we went with AskUI and Claude/Anthropic making TARS work with the Computer tool inside `src/askui/tools/computer.py` which is already passed through the `tools` parameter of `act` for, e.g., `src/askui/agent.py`. We can then also remove the dependency of `UiTarsApiHandler` on `AgentOs`.
- Just add `UiTarsHandler` unconditionally to default registry and initialize the settings can client lazily when a user is trying to use it (similar to how we did it, e.g., in `src/askui/models/anthropic/messages_api.py`).

Alternatively, we can also omit UITars from now from the default registry. Users can still use it as a custom model.

README.md
```diff
@@ -159,9 +159,26 @@ You can use Vision Agent with UI-TARS if you provide your own UI-TARS API endpoi
 
 1. Step: Host the model locally or in the cloud. More information about hosting UI-TARS can be found [here](https://github.com/bytedance/UI-TARS?tab=readme-ov-file#deployment).
 
-2. Step: Provide the `TARS_URL` and `TARS_API_KEY` environment variables to Vision Agent.
+2. Step: Provide the `TARS_URL`, `TARS_API_KEY`, and optionally `TARS_MODEL_NAME` environment variables to Vision Agent.
```
adi-wan-askui:
```suggestion
2. Step: Provide the `TARS_URL`, `TARS_API_KEY`, and optionally `TARS_MODEL_NAME` (defaults to `"tgi"`) environment variables to Vision Agent.
```

or 


```suggestion
2. Step: Provide the `TARS_URL`, `TARS_API_KEY`, and optionally `TARS_MODEL_NAME` (defaults to `"ui-tars"`) environment variables to Vision Agent.
```

depending on the anwer above.

README.md
```diff
@@ -159,9 +159,26 @@ You can use Vision Agent with UI-TARS if you provide your own UI-TARS API endpoi
...
+
+**Example Code:**
+```python
+from askui import VisionAgent
+
+# Set environment variables (or set them in your shell)
+import os
+os.environ["TARS_URL"] = "http://your-tars-endpoint.com"
+os.environ["TARS_API_KEY"] = "your-tars-api-key"
+os.environ["TARS_MODEL_NAME"] = "your-model-name"  # Optional, defaults to "ui-tars"
```
adi-wan-askui: WDYT about omitting this? I fear that this could lead to user's committing their api keys to their Git repos.

adi-wan-askui (changes requested):
Thx so much for your contribution @stephenkopylov :) 

We neglected the UI Tars support quite a bit so it is great to see you contributing here.

src/askui/models/ui_tars_ep/ui_tars_api.py
stephenkopylov: If you think this change might break something, we can leave it as “tgi”.

However, the integration is already slightly broken (setting the TARS_URL and TARS_API_KEY variables didn’t work as documented), so at this stage I think we can change this variable freely and:

- use the most popular model's name (“ByteDance-Seed/UI-TARS-1.5-7B” - when using something like vLLM or LM Studio you'll see this in /v1/models)
- make the TARS_MODEL_NAME variable mandatory for the integration to run (I think this is the cleanest approach) 
- or simply leave “tgi”.

Frankly, I don’t understand why it’s named “tgi” by default :(

README.md
stephenkopylov: Agreed - I’ll adjust this slightly

src/askui/models/model_router.py
stephenkopylov: I’ve moved `src/askui/models/ui_tars_ep/ui_tars_api.py` from using `agent_os` to `computer`.

We still need `from askui.models.shared.tools import Tool` inside the model

`UiTarsHandler` now always exists in the default registry and requires environment variables to be instantiated on the first call.

README.md
stephenkopylov: Done

src/askui/models/ui_tars_ep/ui_tars_api.py
adi-wan-askui: I think we had `"tgi"` because we used a self-hosted tars and called it that way... But not 100% sure anymore. 

Thanks for the detailed analysis :) 

I would leave the decision of what to do here up to you. Preference on _cleanest approach_.

src/askui/models/ui_tars_ep/ui_tars_api.py
```diff
@@ -213,8 +214,21 @@ def act(
...
+        computer_tool = None
+        if tools:
+            for tool in tools:
+                if tool.name == "computer":
+                    computer_tool = tool
+                    break
+
+        if computer_tool is None:
+            error_msg = "Computer tool is required for UI-TARS act() method"
+            raise ValueError(error_msg)
```
adi-wan-askui, line 229: Nicely solved :)

src/askui/models/ui_tars_ep/ui_tars_api.py
```diff
@@ -321,21 +335,20 @@ def execute_act(self, message_history: list[dict[str, Any]]) -> None:
             message_history.append(
                 {"role": "user", "content": [{"type": "text", "text": str(e)}]}
             )
-            self.execute_act(message_history)
+            self.execute_act(message_history, computer_tool)
             return
 
         action = message.parsed_action
         if action.action_type == "click":
```
adi-wan-askui, line 343: Looking at it, I think a tars specific computer model would make sense in the future. Also some actions still need proper implementation. Ideally optional so that UI TARS can alos be used just for chatting without tooling. But I think the current solution is great for now and already a big step forward.

src/askui/models/model_router.py
adi-wan-askui: LGTM 🚀

src/askui/models/ui_tars_ep/ui_tars_api.py
stephenkopylov: Hi @adi-wan-askui
Done - `TARS_MODEL_NAME` is required now :)

danyalxahid-askui:
Thanks again for your contribution @stephenkopylov :)

danyalxahid-askui (approved):

PR #102, opened 2025-07-24
refactor: enhance ComputerToolBase functionality and add coordinate s…

danyalxahid-askui:
…caling

- Added `_get_mouse_position_scaled` method to `ComputerToolBase` for retrieving and scaling mouse position.
- Updated `action` method to return scaled mouse position.
- Introduced `scale_coordinates_with_padding` function in `image_utils.py` for scaling coordinates with padding.
- Cleaned up import statements for better organization and readability.

### DIFF
--- a/src/askui/utils/image_utils.py
+++ b/src/askui/utils/image_utils.py
@@ -7,9 +7,8 @@ from io import BytesIO
 from pathlib import Path
 from typing import Any, Literal, Tuple, Union
 
-from PIL import Image
+from PIL import Image, ImageDraw, ImageOps, UnidentifiedImageError
 from PIL import Image as PILImage
-from PIL import ImageDraw, ImageOps, UnidentifiedImageError
 from pydantic import ConfigDict, RootModel, field_validator
 
 # Regex to capture any kind of valid base64 data url (with optional media type and ;base64)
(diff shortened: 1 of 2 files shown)


src/askui/utils/image_utils.py
```diff
@@ -190,6 +190,55 @@ def scale_image_with_padding(
...
+        y (float): The y-coordinate in the original image.
+        original_width (int): The width of the original image.
+        original_height (int): The height of the original image.
+        max_width (int): The maximum width of the output scaled and padded image.
+        max_height (int): The maximum height of the output scaled and padded image.
+
+    Returns:
+        Tuple[float, float]: A tuple of (scaled_x, scaled_y) coordinates
+        in the padded image.
+    """
```
adi-wan-askui, line 218: I think we should adapt the docstring as it talks about an `image` where there is none 😆

src/askui/utils/image_utils.py
```diff
@@ -190,6 +190,55 @@ def scale_image_with_padding(
...
+    pad_left = (max_width - scaled_width) // 2
+    pad_top = (max_height - scaled_height) // 2
+
+    scaled_x = x * scale_factor + pad_left
+    scaled_y = y * scale_factor + pad_top
+
+    if scaled_x < 0 or scaled_y < 0 or scaled_x > max_width or scaled_y > max_height:
+        error_msg = "Coordinates are outside the padded image area"
+        raise ValueError(error_msg)
+
```
adi-wan-askui, line 238: Can this even happen? Looks to me like this can be removed.

src/askui/utils/image_utils.py
```diff
@@ -190,6 +190,55 @@ def scale_image_with_padding(
     )
 
 
+def scale_coordinates_with_padding(
```
adi-wan-askui, line 193: Can we reuse this inside the `scale_image_with_padding`?

adi-wan-askui (changes requested):
Good work :)

PR #105, opened 2025-07-28
feat: set askui controller path directly

adi-wan-askui:
- intro new `ASKUI_CONTROLLER_PATH` environment variable
- enables to use custom controller path without relying on that
  AskUI (Suite) Installer ran


onur-askui (approved):
LGTM 🚀 

Edit: I think, I caused a conflict. Sorry -_-

PR #108, opened 2025-07-30
feat(models)!: based get model on google genai api

adi-wan-askui:
- `askui` model now uses `gemini-2.5-flash` as default model
   falling back to original `askui` model
   (Inference API's VQA endpoint) if the Google
   GenAI API fails, e.g., because of missing support of
   schema or for unknown reason. For example,
   Google GenAI API does not support recursive schemas
   at the moment.
- `askui/gemini-2.5-flash` and `askui/gemini-2.5-pro` are
  now supported as model choices.
- We are using an AskUI hosted VertexAI proxy for the
  Google GenAI API to ensure compliance, e.g.,
  only EU hosting.

BREAKING CHANGE:
- The `askui`/default model for `AgentBase.get()` changed.


src/askui/models/askui/google_genai_api.py
```diff
@@ -0,0 +1,92 @@
...
+        return model_choice[ASKUI_MODEL_CHOICE_PREFIX_LEN:]
+    return model_choice
+
+
+class AskUiGoogleGenAiApi(GetModel):
+    def __init__(self, settings: AskUiInferenceApiSettings | None = None) -> None:
+        self._settings = settings or AskUiInferenceApiSettings()
+        self._client = genai.Client(
+            vertexai=True,
+            api_key="Necessary",
```
onur-askui: Should we indicate somehow that this is our dummy value so that nobody gets tempted to pass a api key? Is there a different reason for this line than the lib requiring it?

adi-wan-askui: No, there is no different reason other than the lib requiring it. I thought that `Necessary` was enough. It can currently not be overridden by the consumers so I see the contributors as the only people that need to understand this. Do you have an idea for a better, more descriptive default? WDYT of `"DummyValueRequiredByGenaiClient"`? :)

onur-askui: sgtm 👍

PR #113, opened 2025-08-07
fix(retry): update on_exception_types type hint and add support for m…

danyalxahid-askui:
…ultiple exceptions in tests


adi-wan-askui (approved):
Super well done 🚀

PR #114, opened 2025-08-08
Switch multi-tool execution from sequential to parallel with ThreadPoolExecutor

mlikasam-askui:
During testing, running the tools in parallel was up to 9 seconds faster than running them sequentially.
Number of tools: 10
Sequential run: 13.5814 seconds
Parallel (threaded) run: 4.2600 seconds

### DIFF
--- a/src/askui/models/shared/agent.py
+++ b/src/askui/models/shared/agent.py
@@ -165,6 +165,9 @@ class Agent(ActModel):
             for content_block in message.content
             if content_block.type == "tool_use"
         ]
+        if len(tool_use_content_blocks) == 0:
+            return None
+
         content = tool_collection.run(tool_use_content_blocks)
         if len(content) == 0:
             return None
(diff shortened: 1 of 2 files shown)


src/askui/models/shared/agent.py
```diff
@@ -165,6 +165,9 @@ def _use_tools(
             for content_block in message.content
             if content_block.type == "tool_use"
         ]
+        if len(tool_use_content_blocks) == 0:
+            return None
```
adi-wan-askui, line 169: WDYT about removing this? `ThreadPoolExecutor.map()` and, therefore, `ToolCollection.run()` can handle empty list.

adi-wan-askui (changes requested):
Awesome idea and simple implementation, long overdue :) 

The only question I have is, how do we deal with tool calls in the message that are to be called sequentially, example: 

Image

Or can we omit this?

PR #115, opened 2025-08-12
feat(get): support PDF processing and refactor image handling

danyalxahid-askui:
- Updated the `get` method to accept both images and PDFs as sources.
- Renamed the `image` parameter to `source` for clarity.
- Added support for PDF processing in self hosted gemini models.
- Enhanced documentation to reflect the new capabilities and usage examples.
- Introduced a new `PdfSource` class for handling PDF inputs.
- Updated tests to cover new PDF functionality and ensure compatibility with existing image handling.


src/askui/agent_base.py
```diff
@@ -302,13 +310,31 @@ class LinkedListNode(ResponseSchemaBase):
...
-        _image = ImageSource(self._agent_os.screenshot() if image is None else image)
-        self._reporter.add_message("User", f'get: "{query}"', image=_image.root)
+        _source: Source | None = None
+        if source is None:
+            _source = ImageSource(self._agent_os.screenshot())
+        elif isinstance(source, (str, Path)) and Path(source).suffix.lower() == ".pdf":
+            _source = PdfSource(source)
+        else:
+            _source = ImageSource(source)
+
```
adi-wan-askui: would extract this into a function `load_source` similar to `load_image` and instead of relying on the file extension I would go with checking magic numbers using `python-magic` or `filetype.py` as that is more reliable

src/askui/agent_base.py
```diff
@@ -302,13 +310,31 @@ class LinkedListNode(ResponseSchemaBase):
...
+        self._reporter.add_message(
+            "User",
+            f'get: "{query}"',
+            image=_source.root if isinstance(_source, ImageSource) else None,
```
adi-wan-askui, line 339: well done :), maybe we can include the file path in the text if it is available, e.g., `"x/y.pdf"`

src/askui/utils/image_utils.py
```diff
@@ -14,6 +14,32 @@
 # Regex to capture any kind of valid base64 data url (with optional media type and ;base64)
 # e.g., data:image/png;base64,... or data:;base64,... or data:,... or just ,...
 _DATA_URL_GENERIC_RE = re.compile(r"^(?:data:)?[^,]*?,(.*)$", re.DOTALL)
+PDF_MAX_SIZE_BYTES = 20 * 1024 * 1024
+
+
+def load_pdf(source: Union[str, Path]) -> bytes:
```
adi-wan-askui: would move this under `pdf_utils.py` including the other pdf related stuff

tests/e2e/agent/test_get.py
```diff
@@ -43,12 +44,42 @@ def test_get(
...
+        vision_agent.get(
+            "What is in the PDF?",
+            source=path_fixtures_dummy_pdf,
+            model=ModelName.ANTHROPIC__CLAUDE__3_5__SONNET__20241022,
+        )
+
+
+@pytest.mark.parametrize(
+    "model",
+    [
```
adi-wan-askui, line 67: would also add `ModelName.ASKUI` as this should use gemini

src/askui/models/askui/google_genai_api.py
```diff
@@ -112,20 +112,28 @@ def __init__(self, settings: AskUiInferenceApiSettings | None = None) -> None:
...
+            if isinstance(source, ImageSource):
+                part = genai_types.Part.from_bytes(
+                    data=source.to_bytes(),
+                    mime_type="image/png",
+                )
+            elif isinstance(source, PdfSource):
+                part = genai_types.Part.from_bytes(
+                    data=source.root,
+                    mime_type="application/pdf",
+                )
```
adi-wan-askui: would extract this to function

src/askui/models/shared/agent_message_param.py
```diff
@@ -98,6 +109,7 @@ class BetaRedactedThinkingBlock(BaseModel):
     | ToolUseBlockParam
     | BetaThinkingBlock
     | BetaRedactedThinkingBlock
+    | PdfBlockParam
```
adi-wan-askui: Is this already in preparation of the chat?

tests/unit/utils/test_image_utils.py
```diff
@@ -13,11 +14,37 @@
     image_to_base64,
     image_to_data_url,
     load_image,
+    load_pdf,
     scale_coordinates,
     scale_image_to_fit,
 )
 
 
+class TestLoadPdf:
```
adi-wan-askui: I think I would put this into integration tests as we have a dependency on the file system

src/askui/models/shared/agent_message_param.py
danyalxahid-askui: Yes. Removed it for now. Will include it in the next iteration.

src/askui/utils/file_utils.py
```diff
@@ -0,0 +1,33 @@
...
+def load_source(source: Union[str, Path, Image.Image]) -> Source:
+    """Load a source and return appropriate Source object based on file type."""
+
+    if isinstance(source, Image.Image):
+        return ImageSource(source)
+
+    filepath = Path(source)
+    if not filepath.is_file():
+        msg = f"No such file or directory: '{source}'"
+        raise FileNotFoundError(msg)
```
adi-wan-askui: `source` may be a data url. I think this fails for a data url because it assumes a string is a file. I think `load_source` overlaps with `load_image` and `load_pdf`. My idea would be to remove `load_image` and `load_pdf` and integrate the logic into `load_source`. That means we have to move `load_image` out of `ImageSource` and `load_pdf` out of `PdfSource`. I think this makes sense.

tests/e2e/agent/test_get.py
```diff
@@ -77,7 +78,53 @@ def test_get_with_pdf_with_gemini_model(
...
+    path_fixtures_dummy_pdf: pathlib.Path,
+    mocker: MockerFixture,
+) -> None:
+    mocker.patch(
+        "askui.models.askui.google_genai_api.MAX_FILE_SIZE_BYTES",
+        1,
+    )
+
+    # This should raise a ValueError because the default model is Gemini and it falls
+    # back to inference askui which does not support pdfs
```
adi-wan-askui, line 120: Comment was very helpful. But I think we should change implementation to raise the also `PDF file size exceeds the limit` as this would be the reason for it not working.

tests/integration/utils/test_pdf_utils.py
```diff
@@ -0,0 +1,22 @@
+import pathlib
+
+import pytest
+
+from askui.utils.pdf_utils import load_pdf
+
+
+class TestLoadPdf:
+    def test_load_pdf_from_path(self, path_fixtures_dummy_pdf: pathlib.Path) -> None:
```
adi-wan-askui, line 10: Nice tests, data url is missing :) 

See https://gist.github.com/brianguertin/b3ab21de49ac5af81aef58997adc4fac

adi-wan-askui (changes requested):
Close :)

PR #116, opened 2025-08-14
fix: change `ModelName` to a regular class so that (class) constants can be used like regular strings in Python >3.10

adi-wan-askui:
…can be used like regular strings in Python > 3.10

- e.g., fixes issues with using Gemini where, e.g., `"ModelName.GEMINI__2_5__FLASH"` was passed as model id instead of `"gemini-2.5-flash"`


danyalxahid-askui (approved):
lgtm.

Assuming this also solves the issue with boolean response schema for gemini models.

PR #118, opened 2025-08-14
Feat: Add locate all function

mlikasam-askui:
**Breaking changes:**

- The locate function in locator models must now return a list of points instead of a single point.


programminx-askui:
Hello @adi-wan-askui , 

can you take the PR over? @mlikasam-askui  is next week in off.

src/askui/agent_base.py
```diff
@@ -326,8 +326,8 @@ def _locate(
         locator: str | Locator,
         screenshot: Optional[Img] = None,
         model: ModelComposition | str | None = None,
-    ) -> Point:
-        def locate_with_screenshot() -> Point:
+    ) -> list[Point]:
```
adi-wan-askui: The calls of `_locate` are assuming that there is at least one element within the list which is also true judging from the implementation of the models. Can we document that in a docstring using an annotation `Annotation[list[Point], Field(min_len=1)]` as the return type for this here as well as for the `LocateModel`?

Can we also document that depending on the `LocateModel` implementation, the model may only return the first occurence of the element it finds or all occurences?

adi-wan-askui (changes requested):
Well done :) 

Have you checked that all the e2e tests for locate() run through?

adi-wan-askui (approved):

PR #121, opened 2025-08-18
feat(excel): add support for Excel file processing in VisionAgent

danyalxahid-askui:
- Introduced `ExcelSource` class to handle Excel files as input sources.
- Updated `AgentBase` and related APIs to accept `ExcelSource` alongside existing image and PDF sources.
- Implemented error handling for unsupported Excel processing in specific models.
- Added tests for Excel file handling and processing.
- Created utility functions for converting Excel content to markdown format.
- Added dummy Excel file for testing purposes.


pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "protobuf>=6.31.1",
     "google-genai>=1.20.0",
     "filetype>=1.2.0",
+    "markitdown[xls,xlsx]>=0.1.2",
```
adi-wan-askui: I am curious. What made you choose `markitdown` over the alternatives, e.g., docling? Can you write down your evaluation results in our docs?

src/askui/utils/excel_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class ExcelSource(RootModel):
+    """Represents an Excel source that can be read as markdown.
+
+    The class can be initialized with:
+    - A file path (str or pathlib.Path)
+
+    Attributes:
+        root (bytes): The underlying Excel bytes.
```
adi-wan-askui: Can also be a path

src/askui/models/anthropic/messages_api.py
```diff
@@ -242,8 +243,10 @@ def get(
         response_schema: Type[ResponseSchema] | None,
         model_choice: str,
     ) -> ResponseSchema | str:
-        if isinstance(source, PdfSource):
-            err_msg = f"PDF processing is not supported for the model {model_choice}"
+        if isinstance(source, (PdfSource, ExcelSource)):
+            err_msg = (
+                f"PDF or Excel processing is not supported for the model {model_choice}"
+            )
             raise NotImplementedError(err_msg)
```
adi-wan-askui: I think it was a good choice to go that route for now. Later we can use `markitdown` to actually support these kinds of files.

src/askui/models/askui/google_genai_api.py
```diff
@@ -185,6 +186,16 @@ def _create_genai_part_from_source(self, source: Source) -> genai_types.Part:
...
+        if isinstance(source, ExcelSource):
+            with source.reader as r:
+                data = r.read()
+                if len(data) > MAX_FILE_SIZE_BYTES:
+                    _err_msg = (
+                        "Excel file size exceeds the limit of "
+                        f"{MAX_FILE_SIZE_BYTES} bytes."
+                    )
+                    raise ValueError(_err_msg)
+                return genai_types.Part.from_text(text=data.decode())
```
adi-wan-askui: Here there is not really a `MAX_FILE_SIZE_BYTES` as this is passed as simple text content to the model. The main limit here from my perspective is the context window but it could be worth taking another look into the Gemini docs. 

Also I don't like that we encode, just to decode it again. Maybe we should switch to a different pattern where we don't call it `text_reader` and return something like `TextIO` (see Python `io`). Then, we could later also have this for pdf.  But let's think about this in the future.

tests/e2e/agent/test_get.py
```diff
@@ -125,6 +125,47 @@ def test_get_with_pdf_too_large_with_default_model(
...
+
+def test_get_with_xlsx_with_default_model_with_chart_data(
+    vision_agent: VisionAgent, path_fixtures_dummy_excel: pathlib.Path
+) -> None:
+    response = vision_agent.get(
+        "What does the chart show?",
+        source=path_fixtures_dummy_excel,
+    )
+    assert isinstance(response, str)
+    assert "count of names" in response.lower()
```
adi-wan-askui: Is this test really reliable? Seems like something different could come out every time I run it.

tests/e2e/agent/test_get.py
```diff
@@ -125,6 +125,47 @@ def test_get_with_pdf_too_large_with_default_model(
...
+
+
```
adi-wan-askui, line 213: Could we have a test where actually extract the whole table using a schema that defines rows as an array of columns that in turn are an array of strings, e.g., something similar to 

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/schemas/table-object.json",
  "title": "Table as a Single Object",
  "description": "A compact schema for a table represented as an object with 'headers' and 'rows' properties. The number of items in each row array should match the number of headers.",
  "type": "object",
  "properties": {
    "headers": {
      "description": "An array of unique strings for the column headers.",
      "type": "array",
      "items": { "type": "string" },
      "minItems": 1,
      "uniqueItems": true
    },
    "rows": {
      "description": "An array of arrays, where each inner array is a row.",
      "type": "array",
      "items": {
        "type": "array",
        "items": {
          "type": [ "string", "number", "boolean", "null" ]
        }
      }
    }
  },
  "required": [ "headers", "rows" ]
}
```

adi-wan-askui (changes requested):
Nice work :)

pyproject.toml
danyalxahid-askui: Added inside docs.

tests/e2e/agent/test_get.py
danyalxahid-askui: improved it to get data for a specific user.

PR #122, opened 2025-08-18
feat: mcp tool support

adi-wan-askui:
Adds basic support for using MCP tools to both chat and `BaseAgent.act()` by specifying `fastmcp.Client` as a parameter to `ToolCollection`.


docs/mcp.md
```diff
@@ -0,0 +1,167 @@
...
+from askui.agent import VisionAgent
+from askui.models.shared.agent_message_param import MessageParam
+from askui.models.shared.agent_on_message_cb import OnMessageCbParam
+from askui.models.shared.tools import ToolCollection
+from askui.tools.mcp.config import StdioMCPServer
+
+# Create MCP configuration
+mcp_config = MCPConfig(
+    mcpServers={
+        # Make sure to use our patch of StdioMCPServer as we don't support the official one
```
danyalxahid-askui, line 41: Is there a specific reason for that? For the keep_alive functionality?

Is this specifically because after every askui run the subprocesses are also closed by 'StdioTransport'?

src/askui/models/shared/tools.py
```diff
@@ -108,14 +142,49 @@ class ToolCollection:
...
-    def __init__(self, tools: list[Tool] | None = None) -> None:
+    def __init__(
+        self,
+        tools: list[Tool] | None = None,
+        mcp_client: Client[ClientTransportT] | None = None,
+    ) -> None:
         _tools = tools or []
         self._tool_map = {tool.to_params()["name"]: tool for tool in _tools}
+        self._mcp_client = mcp_client
+        self._mcp_tools_cache: dict[str, McpTool] | None = None
```
danyalxahid-askui:
```suggestion
```

pyproject.toml
```diff
@@ -26,6 +27,7 @@ dependencies = [
     "protobuf>=6.31.1",
     "google-genai>=1.20.0",
     "filetype>=1.2.0",
+    "asyncer>=0.0.8",
```
danyalxahid-askui: Would set it to the exact version of this library as suggested by the [author](https://pypi.org/project/asyncer/).

docs/mcp.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Caveats of Using MCP with Chat
+
+When using MCP through the Chat API, consider these limitations:
+
+- **Configuration Limit**: Maximum of 100 MCP configurations allowed
+- **Universal Availability**: All servers are currently passed to all available agents
+- **No Filtering**: No way to filter servers or tools for specific use cases
+- **Tool Name Prefixing**: Tool names are automatically prefixed with the MCP config ID
+- **Server Availability**: When a server is not available (Chat API cannot connect), it is silently ignored
```
danyalxahid-askui, line 136: Is it also viewable in the history? E.g. chat that this tool failed?

docs/mcp.md
```diff
@@ -0,0 +1,167 @@
...
+- **Limited Tool Response Content Types**: Only text and images (JPEG, PNG, GIF, WebP) are supported
+- **Complexity Limits**: Tools are limited in number and complexity by the model's context window
+
+### With Chat
+
+To use MCP servers or, more specifically, the tools they provide with the Chat (API), you need to create MCP configs. All agents are going to have access to the servers specified within the configs if it possible to connect to them.
+
+#### Creating MCP Configs
+
+An MCP configuration can be created with either stdio or remote server similar with what is used when constructing the `MCPConfig` in the library example above
```
danyalxahid-askui, line 98: Can we use this to also read and register MCP's from the json files automatically?

docs/mcp.md
```diff
@@ -0,0 +1,167 @@
...
+
+```bash
```
danyalxahid-askui, line 100: This implementation can be used inside hub to setup configs with vision agent mcps? and on startup those mcp configs are automatically registered. Correct?

danyalxahid-askui: For future inside hub we can guide the user too add other configurations to edit aswell e.g. "authentication" for RemoteMCPServer and "env" inside StdioMCPServer.

Do you see it like that?

src/askui/chat/api/mcp_configs/service.py
```diff
@@ -0,0 +1,103 @@
...
+        self,
+        query: ListQuery,
+    ) -> ListResponse[McpConfig]:
+        mcp_config_paths = list_resource_paths(self._mcp_configs_dir, query)
+        mcp_configs: list[McpConfig] = []
+        for f in mcp_config_paths:
+            try:
+                mcp_config = McpConfig.model_validate_json(f.read_text())
+                mcp_configs.append(mcp_config)
+            except ValidationError:  # noqa: PERF203
```
danyalxahid-askui, line 41: Is this validation because of some corrupted file which cannot be mapped? Since we also validate via McpConfig inside _save function.

docs/mcp.md
adi-wan-askui: Yes, is because `keep_alive`. Has to do with that we start a new event loop for each new call but the server is expected to be connected to from the same event loop by fastmcp package.

docs/mcp.md
adi-wan-askui: Only in the server logs for now. But if the tool cannot be found, it is also not going to be available in the chat... I create a follow up for showing the status of MCP server inside chat ui (https://askyourui.atlassian.net/browse/CL-1710)

docs/mcp.md
adi-wan-askui: No. You need to call the API for now.

docs/mcp.md
adi-wan-askui: The MCP configs are saved as a resource similar to our other resources and, then, an mcp client is created as soon as a run is started from all available configs.

src/askui/chat/api/mcp_configs/service.py
adi-wan-askui: Exactly, there may be a corrupted file. We cannot trust the file system...

PR #130, opened 2025-08-29
feat(chat): expose mcp for whole chat api under `/mcp/sse`

adi-wan-askui:
- uses sse as transport

### DIFF
--- a/src/askui/chat/api/app.py
+++ b/src/askui/chat/api/app.py
@@ -38,14 +38,32 @@ app = FastAPI(
     dependencies=[SetEnvFromHeadersDep],
 )
 
-# Add CORS middleware
-app.add_middleware(
-    CORSMiddleware,
-    allow_origins=["*"],
-    allow_credentials=True,
-    allow_methods=["*"],
-    allow_headers=["*"],
-)
+
+# Include routers
+v1_router = APIRouter(prefix="/v1")
+v1_router.include_router(assistants_router)
+v1_router.include_router(threads_router)
+v1_router.include_router(messages_router)
+v1_router.include_router(runs_router)
+v1_router.include_router(mcp_configs_router)
+v1_router.include_router(files_router)
+v1_router.include_router(health_router)
+app.include_router(v1_router)
+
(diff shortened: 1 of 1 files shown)


src/askui/chat/api/app.py
```diff
@@ -38,14 +38,32 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:  # noqa: ARG001
...
-app.add_middleware(
-    CORSMiddleware,
-    allow_origins=["*"],
-    allow_credentials=True,
-    allow_methods=["*"],
-    allow_headers=["*"],
-)
+
+# Include routers
+v1_router = APIRouter(prefix="/v1")
```
danyalxahid-askui, line 43: Assuming these are exposed locally we don't need to worry about the authorisation of calling them.

How do you image we handling this if chat api moves to the cloud?

src/askui/chat/api/app.py
```diff
@@ -38,14 +38,32 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:  # noqa: ARG001
...
+v1_router.include_router(assistants_router)
+v1_router.include_router(threads_router)
+v1_router.include_router(messages_router)
+v1_router.include_router(runs_router)
+v1_router.include_router(mcp_configs_router)
+v1_router.include_router(files_router)
```
danyalxahid-askui, line 49: Do we have to worry about the model calling the endpoint "/{file_id}/content" and failing because it cannot process the response?

danyalxahid-askui (commented):
Assuming how these are exposed we don't need to worry about the authorisation of calling them.

src/askui/chat/api/app.py
adi-wan-askui: Exactly, currently, we push this to the future. In the future, I imagine that we handle it similarly to how we handle it in our other services :)

src/askui/chat/api/app.py
adi-wan-askui: For now, this could happen. But this is more the clients responsibility to filter tools and support different kinds of outputs so if a client cannot deal with this, it should filter out the tool.

PR #138, opened 2025-09-05
feat(runner, inference_api): integrate `ModelName` for dynamic model …

danyalxahid-askui:
…selection

- Updated `Runner` to use `ModelName.ASKUI` for the custom agent.
- Using Claude 4 asa a default model for askui inference


danyalxahid-askui:
@adi-wan-askui Will merge this as discussed.

PR #140, opened 2025-09-05
fix(chat)!: hold mcp client sessions

adi-wan-askui:
- sessions are not immediately closed after each call to server but
  instead kept open as expected by mcp protocol
- sessions initiated lazily
- one session per client per workspace (across all chats in workspace)
- add playwwright npx to default seeds --> requiring playwright mcp ⟶ requires node.js to be installed (BREAKING CHANGE)
- no update of mcp tools on server update or update of mcp configs
  --> restart of chat api required
- tools with the same name override each other -->
  tool names need to be unique across all servers
- no limitation on workspaces, mcp servers and tools, resources, prompts
  or open connections/sessions
  --> may lead to unexpected behavior in case of too many
- mcp servers with the same name override each other --> names need to
  be unique
- can easily be extended later to support holding mcp sessions
  across one run if it becomes possible to continue run with new message
  instead of creating new run

### DIFF
--- a/src/askui/chat/api/mcp_clients/manager.py
+++ b/src/askui/chat/api/mcp_clients/manager.py
@@ -54,7 +54,7 @@ class McpClientManager:
 
     async def list_tools(
         self,
-    ) -> list[mcp.types.Tool]:  # TODO Proper cache and parallelization
+    ) -> list[mcp.types.Tool]:
         tools: list[mcp.types.Tool] = []
         for mcp_server_name, mcp_client in self._mcp_clients.items():
             if mcp_server_name not in self._tools:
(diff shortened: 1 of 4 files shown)


src/askui/chat/api/mcp_clients/manager.py
```diff
@@ -0,0 +1,156 @@
...
+    async def __aenter__(self) -> "McpClientManager":
+        return await self.connect()
+
+    async def __aexit__(
+        self,
+        exc_type: Type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: types.TracebackType | None,
+    ) -> None:
+        await self.disconnect()
```
danyalxahid-askui, line 102: Should we also pass in Force as True?

PR #141, opened 2025-09-08
Cl 1679 release 2025 09 rpa crud workflow executions

danyalxahid-askui:


src/askui/chat/api/assistants/seeds.py
```diff
@@ -19,6 +19,10 @@
         "list_displays",
         "set_active_display",
         "retrieve_active_display",
+        "create_workflow_v1_workflows_post",
+        "list_workflows_v1_workflows_get",
+        "modify_workflow_v1_workflows",
+        "create_workflow_execution_v1_workflow_executions",
```
adi-wan-askui, line 25: I would move this to a separate automation agent that similarly to the testing agent that can use other agents (computer agent and webagent). WDYT?

src/askui/chat/api/workflow_executions/service.py
```diff
@@ -0,0 +1,146 @@
...
+    ) -> WorkflowExecution:
+        """Retrieve a specific execution by ID."""
+        try:
+            execution_path = self._get_execution_path(execution_id)
+            execution = WorkflowExecution.model_validate_json(
+                execution_path.read_text()
+            )
+
+            # Check workspace access - allow if workspace_id is None (global access)
+            # or if execution workspace matches or execution has no workspace
```
adi-wan-askui, line 104: Rm?

src/askui/chat/api/workflow_executions/models.py
```diff
@@ -0,0 +1,63 @@
...
+    A workflow execution resource in the chat API.
+
+    Args:
+        id (WorkflowExecutionId): The id of the execution. Must start with the 'exec_' prefix and be
+            followed by one or more alphanumerical characters.
+        object (Literal['execution']): The object type, always 'execution'.
+        created_at (datetime.datetime): The creation time as a datetime.
+        workflow (WorkflowId): The id of the workflow being executed. Must start with the 'wf_' prefix.
+        thread (ThreadId): The id of the thread this execution is associated with. Must start with the 'thread_' prefix.
+        status (ExecutionStatus): The current status of the workflow execution.
```
adi-wan-askui, line 37: Where is the status? Is that not essential for it to know wether it succeeded or not? (see issue)

src/askui/chat/api/workflow_executions/router.py
```diff
@@ -0,0 +1,69 @@
...
+    """Create a new workflow execution."""
+    execution, async_generator = await execution_service.create(
+        workspace_id=askui_workspace, params=params
+    )
+
+    async def _run_async_generator() -> None:
+        async for _ in async_generator:
+            pass
+
+    background_tasks.add_task(_run_async_generator)
```
adi-wan-askui, line 56: I think we should separate the `Execution` and the `Run`. The `Execution` is meant more as a report from my perspective of a run that was executed beforehand. So first the agent runs and then creates an `Execution` to report on that. WDYT?

src/askui/chat/api/workflows/models.py
```diff
@@ -42,19 +46,20 @@ class Workflow(WorkspaceResource):
         description (str): A detailed description of the workflow's purpose and steps.
         tags (list[str], optional): Tags associated with the workflow for filtering or
             categorization. Default is an empty list.
-        workspace_id (WorkspaceId | None, optional): The workspace this workflow belongs to.
+        workspace_id (WorkspaceId, optional): The workspace this workflow belongs to.
     """
 
     id: WorkflowId
     object: Literal["workflow"] = "workflow"
+    assistant_id: AssistantId
```
adi-wan-askui, line 54: Why do we need the assistants id here? I would expect the workflow to be independent of the assistant for now. (Btw.: this is also not documented above but not necessary if we decide to remove it)

PR #142, opened 2025-09-08
Refactor/replace hard coded agents with dynamic assistants

adi-wan-askui:


pyproject.toml
```diff
@@ -218,7 +218,7 @@ android = [
     "pure-python-adb>=0.3.0.dev0"
 ]
 chat = [
-    "askui[android,pynput,web]",
+    "askui[android,web]",
```
danyalxahid-askui, line 221: I dont think this change is reflected inside the pdm.lock since they still have the "chat" group inside some packages used by pynput e.g. "six", ""python-xlib"", "pynput"

src/askui/__init__.py
```diff
@@ -2,6 +2,10 @@
 
 __version__ = "0.15.1"
 
+import os
+
+os.environ["FASTMCP_EXPERIMENTAL_ENABLE_NEW_OPENAPI_PARSER"] = "true"
```
danyalxahid-askui, line 7: nice

src/askui/chat/api/app.py
```diff
@@ -57,8 +59,11 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:  # noqa: ARG001
 v1_router.include_router(health_router)
 app.include_router(v1_router)
 
+
 mcp = FastMCP.from_fastapi(app=app, name="AskUI Chat MCP")
 mcp.mount(computer_mcp)
+mcp.mount(testing_mcp)
```
danyalxahid-askui, line 65: Are all of these not made available already inside when we initialize Fastmcp.from_fastapi as most of them are endpoints which also call the same service function which this tool calls?

src/askui/chat/api/assistants/service.py
```diff
@@ -102,6 +106,14 @@ def _save(self, assistant: Assistant, new: bool = False) -> None:
 
     def seed(self) -> None:
         """Seed the assistant service with default assistants."""
+        while True:
```
danyalxahid-askui, line 106: This will remove all previously created assistants. Including the one created by the user.

But since we only use creating an assistant via api and not via Hub this breaking change is acceptable right?

src/askui/chat/api/mcp_configs/service.py
```diff
@@ -143,6 +146,14 @@ def _save(self, mcp_config: McpConfig, new: bool = False) -> None:
 
     def seed(self) -> None:
         """Seed the MCP configuration service with default MCP configurations."""
+        while True:
+            list_response = self.list_(
```
danyalxahid-askui, line 147: Same asassitants.

src/askui/chat/api/runs/runner/runner.py
```diff
@@ -87,156 +90,27 @@ def __init__(
...
-        self, send_stream: ObjectStream[Events], mcp_client: McpClient | None
-    ) -> None:
-        await self._run_agent(
-            agent_type="web_testing",
-            send_stream=send_stream,
-            mcp_client=mcp_client,
-        )
+    def _build_system(self) -> str:
+        base_system = self._assistant.system or ""
+        metadata = {
```
danyalxahid-askui, line 102: Nice

src/askui/prompts/system.py
```diff
@@ -97,29 +97,162 @@
...
+Your final output should only include the content within the <response> and <next_action> tags. Do not include any other tags or internal thought processes in your final output.
+"""
+
+
+ORCHESTRATOR_AGENT_SYSTEM_PROMPT = """
+You are an AI agent called "Orchestrator" with the ID "asst_68ac2c4edc4b2f27faa5a258". Your primary role is to perform high-level planning and management of all tasks involved in responding to a given prompt. For simple prompts, you will respond directly. For more complex, you will delegate and route the execution of these tasks to other specialized agents.
+
+You have the following tools at your disposal:
+
+1. list_assistants_v1_assistants_get
```
danyalxahid-askui, line 201: We are by default giving all of this access to the agent.

Connects to the same as shoudl we go the mcp api endpoint exposing route or the tooling specific route?

pyproject.toml
adi-wan-askui: Great catch :)

src/askui/chat/api/app.py
adi-wan-askui: Nope, they are not, none of these tools are endpoints but rather only tools which is why `FastMCP.from_fastapi` does not include them. In the future we may decide to expose some of them as endpoints as well but this is out of scope for now.

src/askui/chat/api/assistants/service.py
adi-wan-askui: This is meant to only remove global/default assistants (that are not linked to a workspace) that were created via seed beforehand and that can neither be created, modified nor deleted by the user using the api. Do you see somewhere that this is not true?

src/askui/chat/api/mcp_configs/service.py
adi-wan-askui: Same as assistants

src/askui/prompts/system.py
adi-wan-askui: Not sure if I understand your question correctly. Are you asking wether we should restrict access to other agents by exposing them through an extra tool? If yes, I think for now the current route is okay. But this is definitely something to think about at a later point.

PR #146, opened 2025-09-12
add agent_os connect

mlikasam-askui:


src/askui/chat/api/runs/runner/runner.py
```diff
@@ -48,6 +48,7 @@ def _get_android_tools() -> list[Tool]:
     )
 
     agent_os = PpadbAgentOs()
+    agent_os.connect()
```
adi-wan-askui, line 51: Where do we disconnect?

mlikasam-askui: Currently, we don’t.
I couldn’t find an appropriate place to handle the disconnection.

adi-wan-askui: We should do this eventually 😆 But okay for now.

PR #150, opened 2025-09-16
Fix/computer type

mlikasam-askui:


adi-wan-askui (changes requested):
Awesome that you fixed the issue :) Do you know what the underlying reason is? Because I don't think it is the overlay with the overlay disabled. 

Still there are some I think that could improve the solution:
Currently. there is no disconnection happening as far as I can see. So connection are going to dangle around until the server is terminated. The computer_connect tool raises an exception if the connection already exists. I don't think exceptions are a good tool to communicate standard conditional logic. I would either not raise an exception at all or have another tool for checking the connection is already open. 

But maybe an easier solution that solves all issues I currently see and also saves context would be to remove the connect tool and just call `AGENT_OS.connect()` and `AGENT_OS.disconnect()` in the `lifespan()` inside `app.py` similar to the following:

```python
@asynccontextmanager
async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:  # noqa: ARG001
    settings = get_settings()
    mcp_config_service = get_mcp_config_service(settings=settings)
    assistant_service = get_assistant_service(settings=settings)
    assistant_service.seed()
    mcp_config_service.seed()
    AGENT_OS.connect()
    try:
        yield
    finally:
        AGENT_OS.disconnect()
        await get_mcp_client_manager_manager(mcp_config_service).disconnect_all(
            force=True
        )
```

Wdyt?

mlikasam-askui:
> Awesome that you fixed the issue :) Do you know what the underlying reason is? Because I don't think it is the overlay with the overlay disabled.
> 
> Still there are some I think that could improve the solution: Currently. there is no disconnection happening as far as I can see. So connection are going to dangle around until the server is terminated. The computer_connect tool raises an exception if the connection already exists. I don't think exceptions are a good tool to communicate standard conditional logic. I would either not raise an exception at all or have another tool for checking the connection is already open.
> 
> But maybe an easier solution that solves all issues I currently see and also saves context would be to remove the connect tool and just call `AGENT_OS.connect()` and `AGENT_OS.disconnect()` in the `lifespan()` inside `app.py` similar to the following:
> 
> ```python
> @asynccontextmanager
> async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:  # noqa: ARG001
>     settings = get_settings()
>     mcp_config_service = get_mcp_config_service(settings=settings)
>     assistant_service = get_assistant_service(settings=settings)
>     assistant_service.seed()
>     mcp_config_service.seed()
>     AGENT_OS.connect()
>     try:
>         yield
>     finally:
>         AGENT_OS.disconnect()
>         await get_mcp_client_manager_manager(mcp_config_service).disconnect_all(
>             force=True
>         )
> ```
> 
> Wdyt?

We could add the connect and disconnect logic to each MCP lifespan instead of the entire app.
However, the main issue is that the lifespan is tied to each request, meaning the controller would be started and terminated for every tool call.
We also need to save the display ID state so it can be restored on startup. 
Ideally, we would use a shutdown hook, but so far i searched, FastMCP does not provide such hooks.
Additionally, the tool URLs would need to be changed because, as it stands, adding this to the lifespan would cause the computer tools server to start on Android tool requests as well.

PR #161, opened 2025-09-26
feat!(logging): switch from app like logging to lib like logging

adi-wan-askui:
- leave configuration of handlers, formatters, log level to the app
  where "askui" package is installed instead of doing that as
  part of agent construction
- use module level logging using `logging.getLogger(__name__)`
- switch from f-strings to lazy formatting for log messages
- use `extra` for adding context to log messages
- prevent "No handlers could be found" warnings by
  adding `logging.NullHandler` to the module level logger

BREAKING CHANGES:
- remove `log_level` parameter from constructors
    - `VisionAgent`
    - `AndroidVisionAgent`
    - `WebVisionAgent`
    - `WebTestingAgent`
    - `AgentBase`
- log format changed


onur-askui (approved):
👍

PR #178, opened 2025-11-02
Wait until

programminx-askui:


src/askui/agent_base.py
```diff
@@ -497,30 +499,143 @@ def locate_all(
...
+        return self._locate(locator=locator, screenshot=screenshot, model=model)
 
-    @telemetry.record_call()
-    @validate_call
+    @telemetry.record_call(exclude={"locator"})
+    @validate_call(config=ConfigDict(arbitrary_types_allowed=True))
     def wait(
         self,
-        sec: Annotated[float, Field(gt=0.0)],
+        until: Annotated[float, Field(gt=0.0)] | str | Locator,
```
danyalxahid-askui, line 508: Assuming we will add about this breaking change in the release notes and upgrade the version to 0.23.0

src/askui/agent_base.py
```diff
@@ -497,30 +499,143 @@ def locate_all(
             "VisionAgent received instruction to locate all matching UI elements %s",
             locator,
         )
-        return self._locate(locator, screenshot, model)
+        return self._locate(locator=locator, screenshot=screenshot, model=model)
 
-    @telemetry.record_call()
-    @validate_call
+    @telemetry.record_call(exclude={"locator"})
```
danyalxahid-askui: Should this be "locator" or "until"?

src/askui/agent_base.py
programminx-askui: if you don't use `agent.wait(sec=0.5)`, then it's not breaking.

programminx-askui: But I  add this to thre release notes

PR #185, opened 2025-11-11
fix anthropic integration + add push and pull to android agent os

mlikasam-askui:


src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -301,3 +302,23 @@ def get_selected_device_infos(self) -> tuple[str, AndroidDisplay]:
         assert self._device is not None
         assert self._selected_display is not None
         return (self._device.serial, self._selected_display)
+
+    def push(self, local_path: str, remote_path: str) -> None:
+        """
+        Push a file to the device.
+        """
+        self._check_if_device_is_selected()
+        assert self._device is not None
```
programminx-askui, line 311: Is the assert needet, when we check_if_device_is_selected in the line before?

src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -301,3 +302,23 @@ def get_selected_device_infos(self) -> tuple[str, AndroidDisplay]:
...
+        if not Path.exists(Path(local_path)):
+            msg = f"Local path {local_path} does not exist"
+            raise RuntimeError(msg)
```
programminx-askui: Please use a custom exeception.`FileNotFoundError`

src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -301,3 +302,23 @@ def get_selected_device_infos(self) -> tuple[str, AndroidDisplay]:
...
+        self._device.push(local_path, remote_path)
+
+    def pull(self, remote_path: str, local_path: str) -> None:
+        """
+        Pull a file from the device.
+        """
+        self._check_if_device_is_selected()
+        assert self._device is not None
```
programminx-askui, line 322: same here

src/askui/android_agent.py
```diff
@@ -314,6 +314,5 @@ def _get_default_settings_for_act(self, model: str) -> ActSettings:
             messages=MessageSettings(
                 system=ANDROID_AGENT_SYSTEM_PROMPT,
                 thinking={"type": "enabled", "budget_tokens": 2048},
-                betas=[],
```
programminx-askui, line 317: Why?

src/askui/tools/android/ppadb_agent_os.py
mlikasam-askui: for the type checker

src/askui/tools/android/ppadb_agent_os.py
mlikasam-askui: for the type checker

src/askui/android_agent.py
mlikasam-askui: This input is not valid. While our backend can handle it, Anthropic will raise an error if it’s set to an empty list.

src/askui/tools/android/ppadb_agent_os.py
programminx-askui: Can we move this to the _check_if_device_is_selected?

src/askui/models/anthropic/models.py
```diff
@@ -37,7 +37,20 @@
...
+        "Your task is to identify the element described by the user "
+        "and output the pixel coordinates of a single click on that element.\n\n"
+        f"- Screenshot resolution: {screen_width}x{screen_height} (width x height)\n"
+        "- Coordinate system origin is the top-left corner.\n"
+        "- Output must be a single XML tag in the exact format: <click>x,y</click>\n"
+        "- Use integer pixel values only.\n"
+        f"- Coordinates must be within bounds: 0 <= x < {screen_width}, "
+        f"0 <= y < {screen_height}.\n"
+        "- Click near the visual center of the target element.\n"
+        "- Do not include any explanation or additional text."
```
adi-wan-askui, line 52: @mlikasam-askui What about adding a fallback instrucntion for dealing with the case something cannot be located, e.g., something like `"- If you cannot identify the element, return (only exception where you may be out of bounds): <click>-1,-1</click>"`?

src/askui/tools/askui/askui_controller.py
```diff
@@ -79,7 +79,12 @@ def _start_process(
         commands = [str(path)]
         if args:
             commands.extend(args.split())
-        self._process = subprocess.Popen(commands)
+        if not logger.isEnabledFor(logging.DEBUG):
+            self._process = subprocess.Popen(
+                commands, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+        else:
+            self._process = subprocess.Popen(commands)
```
adi-wan-askui, line 87: Nice solution :)

src/askui/agent_base.py
```diff
@@ -48,12 +48,12 @@ class AgentBaseSettings(BaseSettings):
         env_nested_delimiter="__",
         extra="ignore",
     )
-    m: ModelChoice | ModelComposition | str | None = Field(default=None, alias="MODEL")
-    m_provider: str | None = Field(default=None, alias="MODEL_PROVIDER")
+    model: ModelChoice | ModelComposition | str | None = Field(default=None)
+    model_provider: str | None = Field(default=None)
```
adi-wan-askui, line 52: Why this change? I think this is now going to raise warnings as `model_` is a protected namespace? See https://docs.pydantic.dev/2.0/usage/model_config/#protected-namespaces

src/askui/android_agent.py
adi-wan-askui: Good catch @mlikasam-askui  :)

src/askui/agent_base.py
```diff
@@ -157,11 +161,13 @@ def _get_model(
...
-        if model is None:
-            return self._model[type_]
+        if model is None and self.model_name_selected_by_user:
+            return self.model_name_selected_by_user
+
         if isinstance(model, ModelComposition):
             return model
-        return f"{self._model_provider_prefix}{model}"
+
+        return self._model[type_]
```
adi-wan-askui, line 170: Good catch

src/askui/agent_base.py
mlikasam-askui: The environment variables were not working for me

src/askui/models/anthropic/models.py
mlikasam-askui: I like it, that would be a nice feature.

PR #186, opened 2025-11-11
feat(chat): add parent_id to messages for tree navigation

danyalxahid-askui:
- Introduced `parent_id` field in `Message` models to establish parent-child relationships between messages.
- Updated `ChatHistoryManager`, `MessageService`, and related components to support tree navigation based on `parent_id`.
- Enhanced message listing functionality to allow retrieval of messages in a tree path.
- Added migration script to update existing messages with appropriate `parent_id` values.
- Updated relevant tests and documentation to reflect these changes.

### DIFF
--- a/src/askui/chat/api/messages/router.py
+++ b/src/askui/chat/api/messages/router.py
@@ -24,6 +24,38 @@ def list_messages(
     )
 
 
+@router.get("/{message_id}/siblings")
+def list_siblings(
+    askui_workspace: Annotated[WorkspaceId, Header()],
+    thread_id: ThreadId,
+    message_id: MessageId,
+    message_service: MessageService = MessageServiceDep,
+) -> list[Message]:
+    """List all sibling messages for a given message.
+
+    Sibling messages are messages that share the same `parent_id` as the specified message.
+    The specified message itself is included in the results.
+    Results are sorted by ID (chronological order, as IDs are BSON-based).
+
+    Args:
+        askui_workspace (WorkspaceId): The workspace ID from header.
+        thread_id (ThreadId): The thread ID.
+        message_id (MessageId): The message ID to find siblings for.
+        message_service (MessageService): The message service dependency.
+
+    Returns:
+        list[Message]: List of sibling messages sorted by ID.
+
+    Raises:
+        NotFoundError: If the specified message does not exist.
+    """
+    return message_service.list_siblings(
+        workspace_id=askui_workspace,
+        thread_id=thread_id,
+        message_id=message_id,
+    )
+
+
 @router.post("", status_code=status.HTTP_201_CREATED)
 async def create_message(
     askui_workspace: Annotated[WorkspaceId, Header()],
--- a/src/askui/chat/api/messages/service.py
+++ b/src/askui/chat/api/messages/service.py
@@ -437,3 +437,55 @@ class MessageService:
         message_orm = self._find_by_id(workspace_id, thread_id, message_id)
         self._session.delete(message_orm)
         self._session.commit()
+
+    def list_siblings(
+        self,
+        workspace_id: WorkspaceId,
+        thread_id: ThreadId,
+        message_id: MessageId,
+    ) -> list[Message]:
+        """List all sibling messages for a given message.
+
+        Sibling messages are messages that share the same `parent_id` as the specified message.
+        The specified message itself is included in the results.
+        Results are sorted by ID (chronological order, as IDs are BSON-based).
+
+        Args:
+            workspace_id (WorkspaceId): The workspace ID.
+            thread_id (ThreadId): The thread ID.
+            message_id (MessageId): The message ID to find siblings for.
+
+        Returns:
+            list[Message]: List of sibling messages sorted by ID.
+
+        Raises:
+            NotFoundError: If the specified message does not exist.
+        """
+        # Query for all sibling messages using a subquery to get parent_id
+        _parent_id_subquery = (
+            select(MessageOrm.parent_id)
+            .filter(
+                MessageOrm.id == message_id,
+                MessageOrm.thread_id == thread_id,
+                MessageOrm.workspace_id == workspace_id,
+            )
+            .scalar_subquery()
+        )
+
+        orms = (
+            self._session.query(MessageOrm)
+            .filter(
+                MessageOrm.parent_id.is_not_distinct_from(_parent_id_subquery),
+                MessageOrm.thread_id == thread_id,
+                MessageOrm.workspace_id == workspace_id,
+            )
+            .order_by(desc(MessageOrm.id))
+            .all()
+        )
+
+        # Validate that the message exists (if no results, message doesn't exist)
+        if not orms:
+            error_msg = f"Message {message_id} not found in thread {thread_id}"
+            raise NotFoundError(error_msg)
+
+        return [orm.to_model() for orm in orms]


src/askui/chat/api/messages/router.py
```diff
@@ -19,6 +19,34 @@ def list_messages(
...
+        thread_id (ThreadId): The thread ID.
+        query (ListQuery): Pagination parameters (after OR before, limit, order).
+        message_service (MessageService): The message service dependency.
+
+    Returns:
+        ListResponse[Message]: Paginated list of messages in the tree path.
+
+    Raises:
+        ValueError: If both `after` and `before` are specified.
+    """
```
adi-wan-askui: How can I now get the messages of other paths (aka branches)?

src/askui/chat/api/messages/models.py
```diff
@@ -80,6 +82,7 @@ class MessageParam(BaseModel):
 class MessageBase(MessageParam):
     assistant_id: AssistantId | None = None
     run_id: RunId | None = None
+    parent_id: MessageId
```
adi-wan-askui: Style Question:

Why not make it nullable instead but use a root message parent id null value?

src/askui/chat/api/messages/service.py
```diff
@@ -88,17 +293,26 @@ def iter(
         self,
         workspace_id: WorkspaceId,
         thread_id: ThreadId,
-        order: ListOrder = "asc",
-        batch_size: int = LIST_LIMIT_DEFAULT,
+        last_message_id: str,
+        order: ListOrder = "desc",
+        batch_size: int = LIST_LIMIT_MAX,
```
adi-wan-askui: Question:

Why the renaming?

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +64,224 @@ def create(
...
+                upper node (root) and leaf_id is the leaf node. Returns `None` if no messages exist
+                in the thread.
+
+        Raises:
+            ValueError: If both `after` and `before` parameters are specified.
+            NotFoundError: If the specified message in `before` or `after` does not exist.
+        """
+        if query.after and query.before:
+            error_msg = "Cannot specify both 'after' and 'before' parameters"
+            raise ValueError(error_msg)
```
adi-wan-askui, line 290: Minor issue:

I think this results in a 500 as there is no handler for value errors, doesn't it? Would add a global error handler in https://github.com/askui/vision-agent/blob/1c5b46d19f5c506af8d80534671cdda04ca70c4e/src/askui/chat/api/app.py for transforming this into 422. WDYT? :)

src/askui/chat/api/messages/service.py
```diff
@@ -88,17 +293,26 @@ def iter(
...
+        # Then iterate through the rest of the path
         has_more = True
-        last_id: str | None = None
+        last_id: str | None = last_message_id
         while has_more:
             list_messages_response = self.list_(
                 workspace_id=workspace_id,
                 thread_id=thread_id,
-                query=ListQuery(limit=batch_size, order=order, after=last_id),
+                query=ListQuery(limit=batch_size, order=order, before=last_id),
```
adi-wan-askui: Potential issue:

Don't we have to revert this back to `after` after having changed the order of messages?

src/askui/chat/api/messages/models.py
```diff
@@ -24,6 +24,8 @@
 from askui.utils.datetime_utils import UnixDatetime, now
 from askui.utils.id_utils import generate_time_ordered_id
 
+ROOT_MESSAGE_PARENT_ID = "msg_000000000000000000000000"
```
adi-wan-askui, line 27: Style:

What about letting the root `parent_id` just be `None` instead?

src/askui/chat/migrations/versions/7b8c9d0e1f2a_add_parent_id_to_messages.py
```diff
@@ -0,0 +1,87 @@
...
+    # Check if parent_id column already exists
+    inspector = sa.inspect(connection)
+    columns = [col["name"] for col in inspector.get_columns("messages")]
+    column_exists = "parent_id" in columns
+
+    # Add parent_id column as nullable first (only if it doesn't exist)
+    if not column_exists:
+        op.add_column(
+            "messages",
+            sa.Column("parent_id", sa.String(27), nullable=True),
```
adi-wan-askui:
```suggestion
            sa.Column("parent_id", sa.String(24), nullable=True),
```

src/askui/chat/migrations/versions/7b8c9d0e1f2a_add_parent_id_to_messages.py
```diff
@@ -0,0 +1,87 @@
...
+    # Get database connection
+    connection = op.get_bind()
+
+    # Check if parent_id column already exists
+    inspector = sa.inspect(connection)
+    columns = [col["name"] for col in inspector.get_columns("messages")]
+    column_exists = "parent_id" in columns
+
+    # Add parent_id column as nullable first (only if it doesn't exist)
+    if not column_exists:
```
adi-wan-askui, line 31: Comment:

Nice that you thought about itempotency :)

src/askui/chat/migrations/versions/7b8c9d0e1f2a_add_parent_id_to_messages.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+
+    # Check if there are any messages with NULL parent_id that need updating
+    null_parent_count_result = connection.execute(
+        sa.text("SELECT COUNT(*) FROM messages WHERE parent_id IS NULL")
+    )
+    null_parent_count = null_parent_count_result.scalar()
+
+    # Only update parent_ids if null_parent_count is not None and greater than 0
+    if null_parent_count is not None and null_parent_count > 0:
```
adi-wan-askui: Question and potential refactoring:

Speed improvement, right, in case the migration already when partially? I would probably remove this to make the code less complex and live with the speed as it is also only a speed improvement in the edge case where all messages have already been updated.

tests/integration/chat/api/test_messages.py
```diff
@@ -9,7 +9,7 @@
 from fastapi.testclient import TestClient
```
adi-wan-askui, line 9: Test:

What do you think about adding some tests for the new functionality? Maybe as a follow up to the role out of the whole chat?

src/askui/chat/api/messages/service.py
```diff
@@ -40,6 +44,23 @@ def _find_by_id(
             raise NotFoundError(error_msg)
         return message_orm
 
+    def get_last_message_id(
```
adi-wan-askui: Style:

```suggestion
    def retrieve_last_message_id(
```

I think `retrieve` is not as overloaded as `get` which could also mean _calculate_, _property access_ etc.

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+                    MessageOrm.thread_id == thread_id,
+                    MessageOrm.workspace_id == workspace_id,
+                )
+                .first()
+            )
+            if parent_message_orm is None:
+                error_msg = (
+                    f"Parent message {params.parent_id} not found in thread {thread_id}"
+                )
+                raise NotFoundError(error_msg)
```
adi-wan-askui, line 259: Comment:

Awesome that you thought about the validation :)

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+                .limit(1)
+            ).scalar_one_or_none()
+
+            # If no descendants found (e.g., query.before points to non-existent message)
+            if leaf_id is None:
+                _msg_id = query.before if query.before else branch_root_id
+                error_msg = f"Message {_msg_id} not found in thread {thread_id}"
+                raise NotFoundError(error_msg)
+
+        return branch_root_id, leaf_id
```
adi-wan-askui: Tests:

I think it would be great to have some automated tests for different cases where we can focus on this specific function or the `list_` function, e.g.,
- simple path (no branching / single leaf)
- tree with 2 paths (branches, 2 leafs)
- `branch_root_id` == `leaf_id` or
- no messages etc.

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
-        """List messages with pagination and filtering."""
-        q = self._session.query(MessageOrm).filter(
-            MessageOrm.thread_id == thread_id,
-            MessageOrm.workspace_id == workspace_id,
+        """List messages in a tree path with pagination and filtering.
+
+        Behavior:
+        - If `before` is provided: Returns path from before that node down to latest leaf in its subtree (excludes the `after` node itself)
+        - If `after` is provided: Returns path from after that node up to root (excludes the `after` node itself)
+        - If neither: Returns main branch (root to latest leaf in entire thread)
```
adi-wan-askui: Doc issue:

Don't we need to update this after having reverted the order?

src/askui/chat/api/messages/service.py
adi-wan-askui: Question and potential issue:

I would have expected it to be exactly the other way around, i.e, that with `after` you set the root id (which may not be an actual root)  and go down to the leaf and with `before` you set the leaf id (which may not be an actual leaf) and go up to get the root?

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+                branch_root_id = self._session.execute(
+                    select(MessageOrm.id)
+                    .filter(
+                        MessageOrm.parent_id == ROOT_MESSAGE_PARENT_ID,
+                        MessageOrm.thread_id == thread_id,
+                        MessageOrm.workspace_id == workspace_id,
+                    )
+                    .order_by(desc(MessageOrm.id))
+                    .limit(1)
+                ).scalar_one_or_none()
```
adi-wan-askui: Refactor: 

Extract to method `_retrieve_latest_root(workspace_id: WorkspaceId, thread_id: thread_id)`

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+                    MessageOrm.id == branch_root_id,
+                )
+                .cte(name="descendants", recursive=True)
+            )
+
+            # Recursively traverse down
+            _descendants_recursive = select(MessageOrm.id, MessageOrm.parent_id).filter(
+                MessageOrm.parent_id == _descendants_cte.c.id,
+            )
+            _descendants_cte = _descendants_cte.union_all(_descendants_recursive)
```
adi-wan-askui: Refactor:

Extract to method `_build_decendents_cte(message_id: MessageId)`

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+
+            # Get the latest leaf (highest ID)
+            leaf_id = self._session.execute(
+                select(_descendants_cte.c.id)
+                .order_by(desc(_descendants_cte.c.id))
+                .limit(1)
+            ).scalar_one_or_none()
```
adi-wan-askui: Refactor:

Extract to method `_retrieve_latest_leaf(message_id: MessageId)` that uses `_build_decendents_cte()`

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+                )
+                .cte(name="ancestors", recursive=True)
+            )
+
+            # Recursively traverse up until we hit ROOT_MESSAGE_PARENT_ID
+            _ancestors_recursive = select(MessageOrm.id, MessageOrm.parent_id).filter(
+                MessageOrm.id == _ancestors_cte.c.parent_id,
+                _ancestors_cte.c.parent_id != ROOT_MESSAGE_PARENT_ID,
+            )
+            _ancestors_cte = _ancestors_cte.union_all(_ancestors_recursive)
```
adi-wan-askui: Refactor:

Extract to method `_build_ancestors_cte(workspace_id: WorkspaceId, thread_id: ThreadId, message_id: MessageId)`

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+
+            # Get the root node (the one with parent_id == ROOT_MESSAGE_PARENT_ID)
+            branch_root_id = self._session.execute(
+                select(MessageOrm.id).filter(
+                    MessageOrm.id.in_(select(_ancestors_cte.c.id)),
+                    MessageOrm.parent_id == ROOT_MESSAGE_PARENT_ID,
+                )
+            ).scalar_one_or_none()
```
adi-wan-askui: Refactor:

Extract to method `_retrieve_root(message_i: MessageId)` that uses `_build_ancestors_cte()`

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+        Raises:
+            ValueError: If both `after` and `before` parameters are specified.
+            NotFoundError: If the specified message in `before` or `after` does not exist.
+        """
+        if query.after and query.before:
+            error_msg = "Cannot specify both 'after' and 'before' parameters"
+            raise ValueError(error_msg)
+
+        branch_root_id: str | None
+        leaf_id: str | None
```
adi-wan-askui: Refactor:

I think the naming might be misleading as these may not be actual root or leaf or the tree. So maybe rather `path_start` and `path_end`.

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+            # Get the root node (the one with parent_id == ROOT_MESSAGE_PARENT_ID)
+            branch_root_id = self._session.execute(
+                select(MessageOrm.id).filter(
+                    MessageOrm.id.in_(select(_ancestors_cte.c.id)),
+                    MessageOrm.parent_id == ROOT_MESSAGE_PARENT_ID,
+                )
+            ).scalar_one_or_none()
+
+            if branch_root_id is None:
+                error_msg = f"Root message not found in thread {thread_id} for message {leaf_id}"
```
adi-wan-askui: UX issue:

I think we can just return that we did not find a message with id `query.after` as this is more understandable for the end user which does not necessarily understand our algorithm. Would also not say anything about the thread as it might not exist.

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+            # Get the latest leaf (highest ID)
+            leaf_id = self._session.execute(
+                select(_descendants_cte.c.id)
+                .order_by(desc(_descendants_cte.c.id))
+                .limit(1)
+            ).scalar_one_or_none()
+
+            # If no descendants found (e.g., query.before points to non-existent message)
+            if leaf_id is None:
+                _msg_id = query.before if query.before else branch_root_id
```
adi-wan-askui: Refactor:


```suggestion
                _msg_id = branch_root_id
```

as we assign `query.before` to `branch_root_id` if provided

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+                error_msg = f"Message {_msg_id} not found in thread {thread_id}"
```
adi-wan-askui: UX issue:

Would just say that query before message with id ... was not found and not mention the thread as the thread might not exist :D

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+            NotFoundError: If the specified message in `before` or `after` does not exist.
+        """
+        # Step 1: Get concrete branch_root_id and leaf_id
+        _endpoints = self._get_path_endpoints(workspace_id, thread_id, query)
+
+        # If no messages exist yet, return empty response
+        if _endpoints is None:
+            return ListResponse(data=[], has_more=False)
+
+        _branch_root_id, _leaf_id = _endpoints
```
adi-wan-askui: Refactor:

Rename as suggested above

src/askui/chat/api/messages/service.py
```diff
@@ -60,23 +81,229 @@ def create(
...
+            # Stop recursion: don't fetch parent of branch_root_id
+            _path_cte.c.id != _branch_root_id,
+        )
+
+        _path_cte = _path_cte.union_all(_path_recursive)
+
+        # Step 3: Fetch messages with pagination and ordering
+        _query = self._session.query(MessageOrm).join(
+            _path_cte, MessageOrm.id == _path_cte.c.id
         )
```
adi-wan-askui: Refactor:

Extract to method `_build_path_query(workspace_id: WorkspaceId, thread_id: ThreadId, query: ListQuery)`

src/askui/chat/api/messages/models.py
danyalxahid-askui: Mostly following the claude api model.

src/askui/chat/api/messages/service.py
danyalxahid-askui: Not renaming but rather using another maximum limit in query  for less db calls during run.

src/askui/chat/api/messages/router.py
adi-wan-askui: Sorry, this was an old message that you already answered ⟶ Not part of the PR but later I can use the down navigation of this here to get another branch

src/askui/chat/api/messages/service.py
danyalxahid-askui: Discussed. Went back to the original order

src/askui/chat/api/messages/service.py
danyalxahid-askui: Discussed. Went back to the original order

src/askui/chat/api/messages/service.py
danyalxahid-askui: Nice catch. Should be fixed now.

src/askui/chat/api/info/router.py
```diff
@@ -1,20 +0,0 @@
-from fastapi import APIRouter, status
```
adi-wan-askui: So now you removed the info router completely which i think is great to focus the PR but what about its integration now? Did we already do it or was it moved to another PR still to be created?

src/askui/chat/migrations/versions/7b8c9d0e1f2a_add_parent_id_to_messages.py
```diff
@@ -33,47 +33,39 @@ def upgrade() -> None:
...
+    thread_ids = [row[0] for row in threads_result]
+
+    # For each thread, set up parent-child relationships
+    for thread_id in thread_ids:
+        # Get all messages in this thread, sorted by ID (which is time-ordered)
+        messages_result = connection.execute(
+            sa.text(
+                "SELECT id FROM messages WHERE thread_id = :thread_id ORDER BY id ASC"
+            ),
+            {"thread_id": thread_id},
```
adi-wan-askui, line 63: Great that you though of using binds :)

src/askui/chat/migrations/versions/7b8c9d0e1f2a_add_parent_id_to_messages.py
```diff
@@ -33,47 +33,39 @@ def upgrade() -> None:
...
-        # For each thread, set up parent-child relationships
-        for thread_id in thread_ids:
-            # Get all messages in this thread, sorted by ID (which is time-ordered)
-            messages_result = connection.execute(
+    # Fetch all threads
+    threads_result = connection.execute(sa.text("SELECT id FROM threads"))
+    thread_ids = [row[0] for row in threads_result]
+
+    # For each thread, set up parent-child relationships
+    for thread_id in thread_ids:
```
adi-wan-askui, line 57: JFYI: I normally think this is an anti-pattern (constructing the list instead of iterating directly over the cursor) as you load all data into memory when constructing the list but as there are unlikely a crazy amount of threads and it is only a list of ids, I think this is fine here. Similar for messages below.

src/askui/chat/migrations/versions/7b8c9d0e1f2a_add_parent_id_to_messages.py
```diff
@@ -33,47 +33,39 @@ def upgrade() -> None:
...
-
-                connection.execute(
-                    sa.text(
-                        "UPDATE messages SET parent_id = :parent_id WHERE id = :message_id"
-                    ),
-                    {"parent_id": parent_id, "message_id": message_id},
-                )
 
     # Make column non-nullable after setting all parent_ids (only if it was just created)
     if not column_exists:
```
adi-wan-askui: This is not really idempotent as now it may fail after creating the column and if we then rerun it, the column is not going to be made non-nullable.

Ideally, we can just remove this condition so in the worst case scenario it just reruns even though the column is already non-nullable. If not, we need to check specifically if the constraint was already set or not instead.

src/askui/chat/migrations/versions/7b8c9d0e1f2a_add_parent_id_to_messages.py
adi-wan-askui: What about adding foreign key relationship (and with that indexing that that column)? Seems to be missing. You can check the other migrations, e.g., when creating messages table if you need some examples of how to do that.

tests/integration/chat/api/test_message_service.py
```diff
@@ -0,0 +1,464 @@
+"""Unit tests for the MessageService."""
```
adi-wan-askui, line 1: JFYI: Skipped reviewing this in depth but looked like that this provides basic testing of the changes - like :)

src/askui/chat/api/messages/orms.py
```diff
@@ -45,7 +45,6 @@ class MessageOrm(Base):
     )
     parent_id: Mapped[str] = mapped_column(
         MessageId,
-        ForeignKey("messages.id", ondelete="CASCADE"),
```
adi-wan-askui: Why remove it? Would readd it. With `ondelete`, I think there are 2 things we can do, either set this to root message id or the cascade delete as before.

src/askui/chat/api/messages/service.py
```diff
@@ -44,6 +44,177 @@ def _find_by_id(
...
+        # No need to filter by thread_id/workspace_id - parent_id relationship ensures correct path
+        _path_recursive = select(MessageOrm.id, MessageOrm.parent_id).filter(
+            MessageOrm.id == _path_cte.c.parent_id,
+            # Stop recursion: don't fetch parent of path_start
+            _path_cte.c.id != path_start,
+        )
+
+        _path_cte = _path_cte.union_all(_path_recursive)
+
+        # Fetch messages with pagination and ordering
```
adi-wan-askui: no pagination and ordering yet ⟶ would adjust docs

src/askui/chat/api/messages/service.py
```diff
@@ -293,26 +406,18 @@ def iter(
         self,
         workspace_id: WorkspaceId,
         thread_id: ThreadId,
-        last_message_id: str,
```
adi-wan-askui: I think the problem with `last_message_id` missing is now that we always go with the latest path/branch in the tree. But we want to with the one the user choose identified by the last message id which is basically the leaf of that path/branch. I think the cleanest way to do that would actually be the approach you took before and adjusting the truncation logic so that it can deal with the iterator being the other way around. The simplest (most efficient) way from my perspective but not 100% memory safe way is to basically retrieve all the message ids in the path (branch) and then iterate through them in an ascending order calling the retrieve function to retrieve it. Even simpler but even less memory safe would be to load all messages into memory immediately the way you did it beforehand and then reverse the order here.

src/askui/chat/api/info/router.py
danyalxahid-askui: Since we made it backwards compatible. There's no need for it right now.

src/askui/chat/api/messages/service.py
danyalxahid-askui: How we do it is we create a new message (doesnt matter if we are editting or rerunning we create a new message) with the same parent id as the main message. And after that we call run endpoint. Since that message is the latest one we will always run that branch.

src/askui/chat/api/info/router.py
adi-wan-askui: 💯

src/askui/chat/api/messages/service.py
adi-wan-askui: Ingenious 😆 Was not obvious to me but makes total sense. Let's keep it like that for now then.

danyalxahid-askui: Removed it from the createrunparam to avoid confusion.

PR #258, opened 2026-04-07
feat(reporting): Resilient composite reporters and lower memory HTML reports

mlikasam-askui:
### Summary
Improves reporting reliability when a single reporter fails, and reduces memory use for large HTML reports that include many screenshots.

### Changes

- ReporterErrorHandler: Wraps a Reporter, catches exceptions from add_message, add_usage_summary, add_cache_execution_statistics, and generate(), logs with logger.exception, and stops calling the wrapped reporter after the first failure so one bad reporter does not break the rest.
- CompositeReporter: Every nested reporter is wrapped with ReporterErrorHandler.
- SimpleHtmlReporter: 
    - Message rows are appended to a temporary file as they arrive instead of storing a full in-memory list.
    -  generate() writes the HTML header, copies the temp file into the report, removes the temp file, and writes the footer. So base64 image data is not all held in memory at once during long runs.

### DIFF
--- a/src/askui/reporting.py
+++ b/src/askui/reporting.py
@@ -325,6 +325,26 @@ class SimpleHtmlReporter(Reporter):
             self._temp_messages_file = self.report_dir / f"_messages_{id(self)}.tmp"
         return self._temp_messages_file
 
+    _MESSAGE_ROW_TEMPLATE = Template(
+        '<tr class="{{ role_lower }}">'
+        '<td class="timestamp">{{ ts_str }} UTC</td>'
+        '<td><span class="role-badge role-{{ role_lower }}">{{ role }}</span></td>'
+        '<td class="content-cell">'
+        "{% if is_json %}"
+        '<div class="json-content">'
+        '<pre><code class="json">{{ content }}</code></pre>'
+        "</div>"
+        "{% else %}"
+        "{{ content }}"
+        "{% endif %}"
+        "{% for image in images %}"
+        '<br><img src="data:image/png;base64,{{ image }}" '
+        'class="message-image" alt="Message image">'
+        "{% endfor %}"
+        "</td>"
+        "</tr>\n"
+    )
+
     def _render_message_row(
         self,
         timestamp: datetime,
@@ -334,30 +354,13 @@ class SimpleHtmlReporter(Reporter):
         images: list[str],
     ) -> str:
         """Render a single conversation message as an HTML table row."""
-        role_lower = role.lower()
-        ts_str = timestamp.strftime("%H:%M:%S.%f")[:-3]
-
-        if is_json:
-            content_html = (
-                f'<div class="json-content">'
-                f'<pre><code class="json">{content}</code></pre>'
-                f"</div>"
-            )
-        else:
-            content_html = content
-
-        images_html = "".join(
-            f'<br><img src="data:image/png;base64,{image}"'
-            f' class="message-image" alt="Message image">'
-            for image in images
-        )
-
-        return (
-            f'<tr class="{role_lower}">'
-            f'<td class="timestamp">{ts_str} UTC</td>'
-            f'<td><span class="role-badge role-{role_lower}">{role}</span></td>'
-            f'<td class="content-cell">{content_html}{images_html}</td>'
-            f"</tr>\n"
+        return self._MESSAGE_ROW_TEMPLATE.render(
+            role_lower=role.lower(),
+            ts_str=timestamp.strftime("%H:%M:%S.%f")[:-3],
+            role=role,
+            content=content,
+            is_json=is_json,
+            images=images,
         )
 
     @override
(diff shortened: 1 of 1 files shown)


src/askui/reporting.py
```diff
@@ -168,7 +236,9 @@ class CompositeReporter(Reporter):
     """
 
     def __init__(self, reporters: list[Reporter] | None = None) -> None:
-        self._reporters = reporters or []
+        self._reporters = [
+            ReporterErrorHandler(reporter) for reporter in reporters or []
```
philipph-askui, line 273: cant we just add the logic directly to the CompositeReporter?

src/askui/reporting.py
```diff
@@ -1135,21 +1229,31 @@ def generate(self) -> None:
...
         self.report_dir.mkdir(parents=True, exist_ok=True)
-        report_path.write_text(html, encoding="utf-8")
+
+        with report_path.open(mode="w", encoding="utf-8") as out:
+            out.write(header_html)
+            if (
+                self._temp_messages_file is not None
+                and self._temp_messages_file.exists()
+            ):
+                with self._temp_messages_file.open(mode="r", encoding="utf-8") as tmp:
```
philipph-askui: what exactly does this do? Will it also delete the temp file?

philipph-askui: What happens if I hit Ctrl+C during the execution or if something crashes? Will I still get a report in the end?

src/askui/reporting.py
```diff
@@ -1092,39 +1216,9 @@ def generate(self) -> None:
...
-                            {% endfor %}
-                        </table>
-                    </div>
-                </div>
-            </body>
-        </html>
         """
 
-        template = Template(template_str)
+        _FOOTER = "            </table>\n                    </div>\n                </div>\n            </body>\n        </html>"
```
philipph-askui, line 1273: what if the content block of the html changes? Then, the number of whitespaces here will be incorrect, no?

src/askui/reporting.py
```diff
@@ -244,29 +318,76 @@ def _format_content(self, content: Union[str, dict[str, Any], list[Any]]) -> str
...
+        is_json: bool,
+        images: list[str],
+    ) -> str:
+        """Render a single conversation message as an HTML table row."""
+        role_lower = role.lower()
+        ts_str = timestamp.strftime("%H:%M:%S.%f")[:-3]
+
+        if is_json:
+            content_html = (
+                f'<div class="json-content">'
```
philipph-askui: can we use Jinja2 again i/o directly rendering html through f-strings?

src/askui/reporting.py
```diff
@@ -156,6 +160,70 @@ def generate(self) -> None:
...
+        self._error_occurred = True
+
+    @override
+    def add_message(
+        self,
+        role: str,
+        content: Union[str, dict[str, Any], list[Any]],
+        image: Optional[Image.Image | list[Image.Image] | AnnotatedImage] = None,
+    ) -> None:
+        if self._error_occurred:
```
philipph-askui, line 219: this means we will only track the first error and stop tracking stuff in the reporter afterwards?

src/askui/reporting.py
```diff
@@ -244,29 +318,76 @@ def _format_content(self, content: Union[str, dict[str, Any], list[Any]]) -> str
             return json.dumps(content, indent=2)
         return str(content)
 
+    def _get_temp_messages_file(self) -> Path:
+        """Return the path to the temporary messages file, creating it if needed."""
+        if self._temp_messages_file is None:
+            self.report_dir.mkdir(parents=True, exist_ok=True)
+            self._temp_messages_file = self.report_dir / f"_messages_{id(self)}.tmp"
```
philipph-askui: use tempfile.mkstemp i/o id(self) for unique temp names

philipph-askui (changes requested):
Did you test this thoroughly, i.e. with edge cases such as crashing / terminating the execution?

src/askui/reporting.py
mlikasam-askui: This was a cleaner implementation. It allows non-faulty reporters to continue working while only failing the faulty reporter.

src/askui/reporting.py
mlikasam-askui: It should maintain the same behavior as before

src/askui/reporting.py
mlikasam-askui: Yes, especially for the faulty reporter, since we don’t control its behavior. 
Users can/have implemented a faulty custom reporter that causes the execution to fail, which is why this change was made

src/askui/reporting.py
mlikasam-askui: yes it will be deleted using the unlink

src/askui/reporting.py
philipph-askui: we could also just add a list (or similar) to the CompositeReporter that tracks which reporter raised an error earlier and skip these during add_message, add_usage_summary and add_cache_execution, no?

I don't really like wrapping a reporter into a ReporterErrorHandler and that in turn into a CompositeReporter.

programminx-askui:
@mlikasam-askui  & @philipph-askui , 

can we finalize this PR?

philipph-askui:
> @mlikasam-askui & @philipph-askui ,
> 
> can we finalize this PR?

@programminx-askui 
we are currently testing it in a project. I propose to wait with merging until we get positive feedback from there.

philipph-askui (approved):

philipph-askui:
We observed an issue with the tempfile during testing. The execution crashed due to a server-side timeout, but then we also got the following error:
**[WinError 32] The process cannot access the file because it is being used by another process.**

The reason seems to be a potential issue in SimpleHtmlReporter._get_temp_messages_file() (reporting.py:330-338). 
tempfile.mkstemp() returns a tuple (fd, name), but only the file name is stored, the file descriptor is discarded via `_`.

On Windows the unclosed file descriptor keeps an OS-level lock on the file. When generate() later tries to delete it with `self._temp_messages_file.unlink()` (line 1272), Windows rejects the call with an error:
[WinError 32] The process cannot access the file because it is being used by another process.

@mlikasam-askui : can you please take care of this?
(e.g. by closing the fd right away (os.close(fd)) or by switching to tempfile.NamedTemporaryFile(delete=False) which gives a proper file handle that can be closed cleanly)

PR #259, opened 2026-04-08
fix(reporting): improve b64 image truncation for html reports

philipph-askui:


src/askui/reporting.py
```diff
@@ -39,7 +39,7 @@ def normalize_to_pil_images(
 
 def truncate_content(
     content: Any,
-    max_string_length: int = 100000,
+    max_string_length: int = 10_000,
```
mlikasam-askui: I’m not a fan of this magic number.

The goal is to make the reporter more readable, but using a fixed truncate_content length can sometimes hide useful information.

I’d suggest applying this only to images. For other types of content, we should find a better approach. For example, using HTML so users can expand and read larger sections when needed

philipph-askui: I agree, this is actually a problem I have solely with images. I will refactor this 👍🏻

mlikasam-askui (approved):

PR #260, opened 2026-04-08
fix: remove incorrect coordinate scaling from mouse scroll deltas (SO…

programminx-askui:
…LENG-332)

Scroll delta values (dx, dy) were being scaled through `_scale_coordinates_back` in `ComputerAgentOsFacade`, treating them as absolute screen coordinates instead of relative amounts. This caused wildly incorrect scroll values sent to AgentOS (e.g., dy=-5 became dy=-189). Also renamed params from x/y to dx/dy across the scroll API for clarity and improved the tool prompt with scroll unit guidance.

### DIFF
--- a/src/askui/tools/computer/mouse_scroll_tool.py
+++ b/src/askui/tools/computer/mouse_scroll_tool.py
@@ -10,9 +10,9 @@ class ComputerMouseScrollTool(ComputerBaseTool):
             name="mouse_scroll",
             description=(
                 "Scroll the mouse wheel at the current position. "
-                "The scroll amount is specified in lines. "
-                "For example, dy=3 scrolls down 3 lines and dy=-3 scrolls up 3 lines. "
-                "Use small values (1-5) for precise scrolling and larger values (10-20) for fast scrolling. "
+                "The scroll amount depends on the operating system. "
+                "On Windows, a value of 150 scrolls approximately one page. "
+                "Start with dy=150 or dy=-150 for a normal scroll and adjust based on the result. "
                 "Note: The actual scroll direction may vary depending on the OS configuration "
                 "(e.g., natural scrolling on macOS reverses the direction). "
                 "If you observe that scrolling moves in the opposite direction than expected, "
@@ -24,7 +24,7 @@ class ComputerMouseScrollTool(ComputerBaseTool):
                     "dx": {
                         "type": "integer",
                         "description": (
-                            "The horizontal scroll amount in lines. "
+                            "The horizontal scroll amount. "
                             "Positive values scroll right, negative values scroll left. "
                             "Use 0 if no horizontal scrolling is needed."
                         ),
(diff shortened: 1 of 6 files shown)


philipph-askui (changes requested):
as far as I see this PR is three minor things:
- renames x, y => dx, dy
- improves tool descriptions
- removes the scaling for mouse scroll

There is a QA issue in the mouse_scroll_tool.py (lines are too long). As soon as this is resolved I am fine with merging it. In fact, I also need this fix for one project 👍🏻

programminx-askui:
Should be fixed now. Can someone of you check which scroll value is fine? Under Windows it's ~150.

philipph-askui:
> Should be fixed now. Can someone of you check which scroll value is fine? Under Windows it's ~150.

With macOS default settings, negative values scroll down not up. 150 is a small but reasonable step

src/askui/tools/computer/mouse_scroll_tool.py
```diff
@@ -8,22 +8,41 @@ class ComputerMouseScrollTool(ComputerBaseTool):
     def __init__(self, agent_os: ComputerAgentOsFacade | None = None) -> None:
         super().__init__(
             name="mouse_scroll",
-            description="Scroll the mouse wheel at the current position.",
+            description=(
+                "Scroll the mouse wheel at the current "
+                "position. The scroll amount depends on "
+                "the operating system. On Windows, a value "
+                "of 150 scrolls approximately one page. "
```
philipph-askui:
```suggestion
                "of 150 scrolls approximately one page. "
                "On MacOS, a value of 150 scrolls approximately "
                "five rows."
```

mlikasam-askui:
The scaling was intentionally added there. The reason is that if the agent wants to scroll by a dy of 5 pixels in a coordinate system where y_max is 20, it needs to be scaled to 50 when the original y_max is 200. Using the same dy value extracted from the scaled image without adjusting it could lead to issues.

philipph-askui:
> The scaling was intentionally added there. The reason is that if the agent wants to scroll by a dy of 5 pixels in a coordinate system where y_max is 20, it needs to be scaled to 50 when the original y_max is 200. Using the same dy value extracted from the scaled image without adjusting it could lead to issues.

does y_max here equal the display's height?

Scrolling seems to work much better for me on a mac without the scaling. In the end I think the crucial part is to provide the agent with some anchor points (i.e. examples) for scrolling amounts in the tool description.

@mlikasam-askui

programminx-askui:
@mlikasam-askui , 

we are mixing up things. The mouse event which is send to the OS describes only the dx & dy in "line height"!? and **not pixel**. This means each application which receives the mouse wheel event can interpret by themself how much pixel it scolls. This means the scroll event is independent of the resolution. So scaling is not necessary.  

-> I tried it with 5, 20, 50 and 150 as default values and felt the 150 fits the best for our use cases. I'm still not 100% sure. 

Addition to this: The scaling implementation had a bug, that the positive values (dx=5) where converted to negative scrolling commands on AgentOs (dx=-180). And this was in the first place the reason to remove it.

mlikasam-askui (approved):

philipph-askui:
@programminx-askui feel free to accept or decline my suggestion above. After that I think we can merge this.

philipph-askui (approved):

PR #261, opened 2026-04-09
feat: add warning to WriteToFileTool if path is longer than Windows max path length (260 characters)

philipph-askui:
if the path is longer than 260 characters, the tool will just return a "No such File or Directory" error.
This is quite hard to debug. Hence I added a warning.

### DIFF
--- a/src/askui/tools/store/universal/write_to_file_tool.py
+++ b/src/askui/tools/store/universal/write_to_file_tool.py
@@ -1,6 +1,9 @@
+import logging
 from pathlib import Path
 
 from askui.models.shared.tools import Tool
+
+logger = logging.getLogger(__name__)
 
 
 class WriteToFileTool(Tool):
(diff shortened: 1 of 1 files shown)


src/askui/tools/store/universal/write_to_file_tool.py
```diff
@@ -118,6 +121,13 @@ def __call__(self, file_path: str, content: str, append: bool) -> str:
             OSError: If the file cannot be written due to filesystem errors.
         """
         absolute_file_path = self._base_dir / file_path
+        if len(str(absolute_file_path)) > 260:
+            logger.warning(
+                "File path exceeds 260 characters (%d characters: '%s')."
+                "This may cause errors on Windows.",
+                len(str(absolute_file_path)),
+                absolute_file_path,
+            )
```
mlikasam-askui, line 130: Why? Should we apply this to all file-related tools, for example the saveScreenshot function?

programminx-askui: Can we have a discussion about the LongPath handling under windows? 

My prefered way would be: 
1. Warn the user at starting (Similar like AskUI Suite)
2. Check file pathes before written and throw an exception with a hint to enable it

philipph-askui: It is not an urgent issue, so I will close this PR for now. I added a ticket to our backlog so that we can discuss that and then implement according to our results.

PR #262, opened 2026-04-10
fix: handle comma-separated coordinates in move_mouse tool

philipph-askui:
The agent quite often passes both x/y coordinates as a single comma-separated string to the `x` parameter (e.g., `x="330, 182"` or `x="54, "`), causing the `int()` conversion to fail.
I added parsing the comma-separated string and extract both coordinates before converting to int to make it more robust

### DIFF
--- a/src/askui/tools/computer/move_mouse_tool.py
+++ b/src/askui/tools/computer/move_mouse_tool.py
@@ -49,7 +49,9 @@ class ComputerMoveMouseTool(ComputerBaseTool):
         _NUMBER_PATTERN = re.compile(r"-?\d+")
         combined = f"{x},{y}"
         numbers = _NUMBER_PATTERN.findall(combined)
-        if len(numbers) < 2:
-            error_msg = f"Could not parse coordinates from x={x!r}, y={y!r}"
+        if not len(numbers) == 2:
+            error_msg = f"""Could not parse mouse_move coordinates from provided
+              parameters x={x}, y={y}. The parameters x and y must be passed as separate
+              integer values!"""
             raise ValueError(error_msg)
         return int(numbers[0]), int(numbers[1])
--- a/src/askui/tools/computer/move_mouse_tool.py
+++ b/src/askui/tools/computer/move_mouse_tool.py
@@ -29,9 +29,16 @@ class ComputerMoveMouseTool(ComputerBaseTool):
         self.is_cacheable = True
 
     def __call__(self, x: int, y: int) -> str:
-        # for some reason, the agent occasionally calls the tool with the coords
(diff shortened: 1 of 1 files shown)


src/askui/tools/computer/move_mouse_tool.py
```diff
@@ -29,9 +29,16 @@ def __init__(self, agent_os: ComputerAgentOsFacade | None = None) -> None:
...
+        # The agent occasionally passes coordinates incorrectly:
+        # 1. As strings instead of ints (e.g., x="330", y="182")
+        # 2. Both coords as a single comma-separated string in x
+        #    (e.g., x="330, 182" or x="330, ")
+        # We handle both cases here.
+        if isinstance(x, str) and "," in x:  # type: ignore[unreachable]
+            parts = [p.strip() for p in x.split(",") if p.strip()]  # type: ignore[unreachable]
+            x = parts[0]
+            if len(parts) > 1:
+                y = parts[1]
```
mlikasam-askui: I don’t see the need for this “fix.” It only addresses one of the potential mistakes the agent might make.
```suggestion
```

mlikasam-askui (changes requested):

src/askui/tools/computer/move_mouse_tool.py
philipph-askui: True, but this is the mistake that I see often (a couple of times in every execution). I don't see any other mistakes often

programminx-askui: I agree with @philipph-askui . We should have tools which "solves" LLM problems. So that the overall Token Consumptions goes down. 

@philipph-askui can we use regex? So it's better readable?

programminx-askui (changes requested):

src/askui/tools/computer/move_mouse_tool.py
mlikasam-askui: @philipph-askui  is often the keyword, though other mistakes can occur as well. I support extending the tool description. 
@programminx-askui , the issue is that modifying the agent’s input may hinder auto-recovery. If the tool isn’t carefully designed, the agent won’t know its input was changed and how it was changed and it might keep trying with the same mistake.

philipph-askui: Fair points. Here is what I would propose
1) We improve the tool description to make this problem less likely
2) If it still occurs, we try to convert to the correct types. If this is possible we can execute the tool. If this is not possible, we raise an error (as if we did not even try to convert).

philipph-askui:
@programminx-askui @mlikasam-askui : I addressed the comments of you both in my last commit

PR #263, opened 2026-04-10
chore: remove get and locate tools from Agent's default tools

philipph-askui:
Get and Locate tools seem to cause issues at times during act commands. I think they should not be added per default.


mlikasam-askui (approved):
<3

programminx-askui:
😭 

But go ahead.

PR #266, opened 2026-04-15
feat: add per-conversation duration to Html reports

philipph-askui:
Adds the duration of each conversation in the per-conversation usage breakdown of the Html reports

image

### DIFF
--- a/src/askui/callbacks/usage_tracking_callback.py
+++ b/src/askui/callbacks/usage_tracking_callback.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from datetime import datetime, timezone
 from typing import TYPE_CHECKING
 
 from opentelemetry import trace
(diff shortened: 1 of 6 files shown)


src/askui/callbacks/usage_tracking_callback.py
```diff
@@ -172,11 +173,22 @@ class StepUsageSummary(UsageSummary):
...
+        step_summaries (list[StepUsageSummary]): Per-step usage summaries.
+        duration_seconds (float | None): Wall-clock duration of the conversation
+            in seconds, measured between `on_conversation_start` and
+            `on_conversation_end`. `None` if duration was not tracked.
+    """
 
     conversation_index: int
     conversation_id: str
     step_summaries: list[StepUsageSummary] = Field(default_factory=list)
+    duration_seconds: float | None = None
```
programminx-askui: Do we not have a times object for duration in Python lib?

src/askui/reporting.py
```diff
@@ -37,6 +37,27 @@ def normalize_to_pil_images(
     return [image]
 
 
+def _format_duration(seconds: float) -> str:
```
programminx-askui, line 43: Would be nice if we have a time object. Then we don't need to implement the format logic by our self

programminx-askui (changes requested):

src/askui/callbacks/usage_tracking_callback.py
programminx-askui: I changed my mind. Please store the start time and end time and generate the duration in the generate function when the html is generated.

Then we have the values also for other information

programminx-askui: And why is this part of the usage tracking? 

Should we change the name?

philipph-askui: both addressed in #e12eb14b25093833b46467064ccc8fcbe61ae994:
- renamed `UsageTrackingCallback` to `ConversationStatisticsCallback`
- stored `started_at` and `ended_at` as datetime objects i/o `duration_seconds` in the `ConversationUsageSummary`

programminx-askui (approved):

PR #269, opened 2026-04-29
feat(agent-os): add keyboard key/modifier aliases and Playwright mappings

mlikasam-askui:


src/askui/tools/agent_os.py
```diff
@@ -23,9 +23,16 @@
...
     "alt",
+    "alt_left",
+    "alt_right",
     "control",
+    "control_left",
+    "control_right",
     "shift",
+    "shift_left",
+    "shift_right",
     "right_shift",
```
mlikasam-askui, line 29: I kept this in place to avoid introducing breaking changes.

src/askui/tools/agent_os.py
```diff
@@ -23,9 +23,16 @@
 
 ModifierKey = Literal[
     "command",
+    "os_modifier",
```
benjamin-askui: I would avoid introducing new names for the same thing for now.

src/askui/tools/agent_os.py
```diff
@@ -36,14 +43,27 @@
     "enter",
     "tab",
     "escape",
+    "insert",
     "up",
+    "arrow_up",
```
benjamin-askui: I would avoid adding duplicate names for the same thing, e.g. `arrow_up` for `up`, `arrow_down` for `down`, etc.

That will just add more confusion than it helps.

src/askui/tools/agent_os.py
```diff
@@ -56,6 +76,49 @@
...
+    "numpad_6",
+    "numpad_7",
+    "numpad_8",
+    "numpad_9",
+    "numpad_+",
+    "numpad_-",
+    "numpad_*",
+    "numpad_/",
+    "numpad_.",
+    "media_volume_mute",
```
benjamin-askui: You can introdce these, but they don't exist on all platforms. For example MacOS doesn't have:
`F21`, `F22`, `F23`, `media_next`, `media_prev`, `media_play_or_pause`, `media_stop`.

Might be confusing for customer if it works on one platform but not on the other.

mlikasam-askui: Do you recommend adding only the numpad keys?

benjamin-askui: I think that would be the best for now, the rest could be added on a need basis.
Especially since we are trying to move to a new system soon anyway.

src/askui/tools/agent_os.py
philipph-askui: In how far would this be a breaking change? Only if these values are stored in some cache files or explicitly written in prompts, no?

philipph-askui (approved):

PR #272, opened 2026-05-04
fix: incorrect typing speed unit of computer type tool

philipph-askui:
the description stated that typing speed was character per minute, in fact, the parameter controls the typing speed in characters per second.

### DIFF
--- a/src/askui/tools/computer/type_tool.py
+++ b/src/askui/tools/computer/type_tool.py
@@ -19,7 +19,7 @@ class ComputerTypeTool(ComputerBaseTool):
                     "typing_speed": {
                         "type": "integer",
                         "description": (
-                            "The speed of typing in characters per minute."
+                            "The speed of typing in characters per second."
                             " Defaults to 50"
                         ),
                         "default": 50,
@@ -35,5 +35,5 @@ class ComputerTypeTool(ComputerBaseTool):
         self.agent_os.type(text, typing_speed)
         return (
             f"Text was typed: {text} with typing speed: "
-            f" {typing_speed} characters per minute."
+            f" {typing_speed} characters per second."
         )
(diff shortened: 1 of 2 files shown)


src/askui/tools/computer/type_tool.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, agent_os: AgentOs | None = None) -> None:
                     "typing_speed": {
                         "type": "integer",
                         "description": (
-                            "The speed of typing in characters per minute."
+                            "The speed of typing in characters per second."
                             " Defaults to 50"
```
mlikasam-askui, line 23: @programminx-askui  should we increase the default value?

philipph-askui: Would be nice, if we could set the default as a a config value so that it can be easily set per project. For example when operating a VNC window, a too high typing speed will lead to some characters being not typed....

mlikasam-askui: Since the agent can modify the value, even changing the default doesn’t guarantee it will be used.
btw, are you sure that the value is per second ?

philipph-askui: The agent usually does not modify the value unless it is specifically prompted to do so. Also, it would make more sense to set it once for these specific scenarios so that the agent does not need to do it for every type call. However, this is a low priority issue in my opinion so I would not add it to this PR for now.

programminx-askui: I have the feeling something is off. And either the typeing speed in the API, the gloabl config is not working or the Agent is randomly changing the speed which leads in chars which are not recognized by the application. 

We need first to do a proper testing, but I woul do it in another bug ticket.

philipph-askui: It seems to work as expected on mac, so this might be a windows issue?

philipph-askui: Anyways, the descriptions should be changed from minutes to seconds (as done in this PR) in order to align with the definition of the API.

PR #274, opened 2026-05-11
Fix: Add messages to reporter if crashing during truncation

philipph-askui:
fix: add messages to reporter before adding them to truncation strategy (so that they show up in case truncation crashes)
fix: report truncation errors to reporter before raising them

### DIFF
--- a/src/askui/models/shared/truncation_strategies.py
+++ b/src/askui/models/shared/truncation_strategies.py
@@ -21,11 +21,11 @@ from askui.models.shared.prompts import SystemPrompt
 from askui.models.shared.token_counter import SimpleTokenCounter
 from askui.models.shared.tools import ToolCollection
 from askui.prompts.truncation import SUMMARIZE_INSTRUCTION_PROMPT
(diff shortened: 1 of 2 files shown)


programminx-askui (approved):

src/askui/models/shared/truncation_strategies.py
```diff
@@ -121,13 +123,21 @@ def _summarize_message_history(
...
+            tools=tools,
+            provider_options=provider_options,
+        )
+    except Exception as e:
+        # catch e.g. BadRequestError
+        if reporter:
+            reporter.add_message(
+                "TruncationStrategy", f"Truncation Failed with error: {e}"
+            )
+        raise
```
programminx-askui, line 140: Should it not be raise e?

philipph-askui: nope: https://docs.astral.sh/ruff/rules/verbose-raise

programminx-askui (approved):

PR #276, opened 2026-05-12
feat: multi Agent OS server support in `AskUiControllerClient`

mlikasam-askui:
## Summary

`MultiComputerTargetAgentOS` can now manage **multiple Agent OS target
computers** (local and/or remote) at once. One is active at a time and receives
the agent's actions; callers can switch at runtime or scope a switch to a `with`
block.

New API:

- `ComputerTarget` (base), `LocalComputerTarget` (replaces the old
  `AskUiControllerServer` — owns the local controller subprocess; auto-detects
  the Windows `AskuiCoreService` and switches to its port),
  `RemoteComputerTarget` (already-running remote controller, no process
  management). Each target owns its own gRPC connection
  (`ComputerTargetConnection`).
- `ComputerTargetPool` — enforces invariants (at most one local, unique session
  GUIDs / `computer_id`s / remote addresses), tracks the active target, and
  orchestrates connect/disconnect. Targets are addressed exclusively by
  `computer_id` (single `dict` lookup; no secondary indices).
- On `MultiComputerTargetAgentOS` and exposed through `ComputerAgentOS` /
  `ComputerAgentOsFacade`: `add_agent_os_target_computer`,
  `describe_agent_os_target_computers` (returns the targets' `repr()` strings,
  not live objects), `get_current_computer_target_id` (returns the `computer_id`
  string of the active target), `switch_agent_os_target_computer`,
  `reset_agent_os_target_computers`, `temporary_select`.
- `PpadbAgentOs` gets a sibling `temporary_select(device_sn)`.
- `ComputerAgent` auto-registers three new `act()` tools so the LLM can drive
  multi-target flows: `ComputerListAgentOsTargetComputersTool`,
  `ComputerSwitchAgentOsTargetComputerTool`,
  `ComputerGetCurrentComputerTargetIdTool`.
- `LocalComputerTarget` and `RemoteComputerTarget` are re-exported from the
  top-level `askui` package.

New unit tests cover the target-computer classes, the pool, the multi-target
client, and the new computer tools. The e2e test was updated to the new
constructor.

## Example

Register targets up front and let `act()` switch between them:

```python
from askui import ComputerAgent
from askui import LocalComputerTarget, RemoteComputerTarget

with ComputerAgent(
    agent_os_target_computers=[
        LocalComputerTarget(computer_id="research-box"),
        RemoteComputerTarget(
            address="192.168.1.42:26000",
            description="Writer box with a text editor open",
            computer_id="writer-box",
        ),
    ],
) as agent:
    agent.act(
        "On research-box, google 'askui' and read the top results. Then switch "
        "to writer-box and write a Markdown summary in the open text editor."
    )
```


src/askui/tools/computer/__init__.py
```diff
@@ -14,12 +16,14 @@
 from .retrieve_active_display_tool import ComputerRetrieveActiveDisplayTool
 from .screenshot_tool import ComputerScreenshotTool
 from .set_active_display_tool import ComputerSetActiveDisplayTool
+from .switch_agent_os_server_tool import ComputerSwitchAgentOsServerTool
```
philipph-askui: shouldn't these 3 new tools also be in experimental?

src/askui/tools/computer/get_active_agent_os_server_tool.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, agent_os: AgentOs | None = None) -> None:
+        super().__init__(
+            name="get_active_agent_os_server",
+            description="""
+                Return the currently active Agent OS server that agent-os actions
+                are routed to.
+            """,
+            agent_os=agent_os,
+        )
+        self.is_cacheable = False
```
philipph-askui: why did you set is_cacheable to False? I think this is absolutely cacheable

src/askui/tools/computer/get_mouse_position_tool.py
```diff
@@ -8,12 +8,20 @@ class ComputerGetMousePositionTool(ComputerBaseTool):
...
             required_tags=[ToolTags.SCALED_AGENT_OS.value],
         )
         self.is_cacheable = True
 
     def __call__(self) -> str:
+        server = self.agent_os.get_active_agent_os_server(report=False)
         cursor_position = self.agent_os.get_mouse_position()
-        return f"Mouse is at position ({cursor_position.x}, {cursor_position.y})."
+        return (
+            f"[Server with id '{server.computer_id}']: Mouse is at position "
```
philipph-askui: I am not happy with the name "Server in general". So far, I was using the term "machine" when telling the agent that it is operating multiple devices.
The name of one machine could then be server. If we use the term server here, this will introduce ambiguity that might cause the currently working stuff to break

src/askui/tools/computer/get_system_info_tool.py
```diff
@@ -29,4 +32,6 @@ def __init__(self, agent_os: AgentOs | None = None) -> None:
         )
 
     def __call__(self) -> str:
-        return str(self.agent_os.get_system_info().model_dump_json())
+        server = self.agent_os.get_active_agent_os_server(report=False)
+        system_info_json = self.agent_os.get_system_info().model_dump_json()
+        return f"[Server with id '{server.computer_id}']: {system_info_json}"
```
philipph-askui: same here (Server ->machine)

src/askui/tools/computer/switch_agent_os_target_computer_tool.py
```diff
@@ -0,0 +1,27 @@
...
+                "type": "object",
+                "properties": {
+                    "computer_id": {
+                        "type": "string",
+                    },
+                },
+                "required": ["computer_id"],
+            },
+            agent_os=agent_os,
+        )
```
philipph-askui, line 25: add self.is_cacheable=True

src/askui/tools/computer/switch_agent_os_target_computer_tool.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __call__(self, computer_id: str) -> str:
```
philipph-askui, line 27: how is this computer_id determined? By the user, e.g. in the init of the agent? Currently, they can set local_machine_name and remote_machine_name

philipph-askui:
tbh: at the moment I really like the concept with 2 dedicated toolsets to operate 2 machines. Still, I see that this here will scale also to even more machines.

Have you checked if the agent understands that it can operate multiple machines and that it calls the switch tool before executing operations if needed?

Further: we need to automatically adjust the system capabilities and device information prompts if multiple agentOS are added.

Also: we should definitely update the docs with this PR to explain these changes here

philipph-askui (changes requested):

src/askui/tools/computer/__init__.py
mlikasam-askui: I see them as part of the default tool list, similar to the Android agent.

src/askui/tools/computer/get_active_agent_os_server_tool.py
mlikasam-askui: The session ID and computer ID are generated dynamically using UUID4 values. Does the cache support handling dynamic values like these?

src/askui/tools/computer/switch_agent_os_target_computer_tool.py
mlikasam-askui: https://github.com/askui/python-sdk/pull/276#discussion_r3246415132

src/askui/tools/computer/switch_agent_os_target_computer_tool.py
mlikasam-askui: It can be set by the user, or it defaults to the session ID, which is a UUID4.

src/askui/models/shared/tools.py
```diff
@@ -534,12 +534,23 @@ def reset_tools(self, tools: list[Tool] | None = None) -> None:
         """Reset the tools in the collection with new tools."""
         self._tools = tools or []
 
-    def get_agent_os_by_tags(self, tags: list[str]) -> AgentOs | AndroidAgentOs:
-        """Get an agent OS by tags."""
+    def get_agent_os_by_tags(
+        self, required_tags: list[str]
+    ) -> AgentOs | AndroidAgentOs:
+        """
+        Find the first registered agent OS whose tags are a superset of
```
programminx-askui:
```suggestion
        Find the first registered AgentOS whose tags are a superset of
```

Please find all occurance. in Strings

src/askui/tools/android/agent_os.py
```diff
@@ -502,3 +504,26 @@ def get_ui_elements(self) -> UIElementCollection:
         Gets the UI elements.
         """
         raise NotImplementedError
+
+    def temporary_select(self, device_sn: str) -> AbstractContextManager[Self]:
```
programminx-askui, line 508:
```suggestion

    @abstractmethod
    def temporary_select(self, device_sn: str) -> AbstractContextManager[Self]:
```

programminx-askui: Ok. 

Why is this function not implmented?

src/askui/tools/android/agent_os_facade.py
```diff
@@ -112,6 +115,15 @@ def set_device_by_serial_number(self, device_sn: str) -> None:
         self._agent_os.set_device_by_serial_number(device_sn)
         self._real_screen_resolution = None
 
+    @contextmanager
+    def temporary_select(self, device_sn: str) -> Iterator[Self]:
+        with self._agent_os.temporary_select(device_sn):
+            self._real_screen_resolution = None
+            try:
+                yield self
+            finally:
```
programminx-askui, line 133: I assume you tested it.

src/askui/tools/android/ppadb_agent_os.py
```diff
@@ -202,6 +205,24 @@ def set_device_by_serial_number(self, device_sn: str) -> None:
         msg = f"Device name {device_sn} not found"
         raise AndroidAgentOsError(msg)
 
+    @contextmanager
+    def temporary_select(self, device_sn: str) -> Iterator[Self]:
```
programminx-askui, line 209: But in general. Why do we need the concept of `temporary_select`? How does the agent should call it?

src/askui/tools/askui/agent_os_server.py
```diff
@@ -0,0 +1,320 @@
...
+
+from typing_extensions import override
+
+from askui.tools.askui.askui_controller_settings import AskUiControllerSettings
+from askui.tools.utils import process_exists, wait_for_port
+
+logger = logging.getLogger(__name__)
+
+
+def _generate_session_guid() -> str:
```
programminx-askui: Do we have already such a function?

src/askui/tools/askui/agent_os_server.py
```diff
@@ -0,0 +1,320 @@
...
+    return "{" + str(uuid.uuid4()) + "}"
+
+
+def _replace_port(address: str, port: int) -> str:
```
programminx-askui: Why do we need a function replace_port?

Can you not set on the parsed URL the port directly? Oterhweise you are throwing away the rest of the url.

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,320 @@
...
+    host = parsed.hostname or "localhost"
+    return f"{host}:{port}"
+
+
+class AgentOsServer:
+    """
+    Base class describing an Agent OS server that the `AskUiControllerClient` can
+    connect to.
+
+    An Agent OS server is the server-side counterpart of the `AgentOs` client
```
programminx-askui:
```suggestion
    An Agent OS server is the server-side counterpart of the `AgentOS` client
```

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,320 @@
...
+            helpers. Must be unique across registered servers. Defaults to the
+            server's `session_guid`.
+    """
+
+    def __init__(
+        self,
+        address: str,
+        description: str,
+        display: int = 1,
+        computer_id: str | None = None,
```
programminx-askui, line 47: Reminder to speak about it

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,320 @@
...
+        return self._display
+
+    @display.setter
+    def display(self, value: int) -> None:
+        self._display = value
+
+    @property
+    def is_local(self) -> bool:
+        """Whether this server represents a locally-managed process."""
+        return False
```
programminx-askui, line 93: ? But maybe I will see it later

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,320 @@
...
+            Service on Windows). Defaults to `False`.
+        discover_service (bool, optional): On Windows, probe for a running
+            ``askuicoreservice`` and, if found, switch the address to port
+            ``26000`` and set `is_service` to `True`. Defaults to `True`.
+        description (str, optional)
+        display (int, optional): Display ID selected for this server. Defaults to `1`.
+    """
+
+    _ASKUI_CORE_SERVICE_NAME = "AskuiCoreService"
+    _ASKUI_CORE_SERVICE_PORT = 26000
```
programminx-askui, line 176: ?

src/askui/tools/askui/agent_os_server.py
```diff
@@ -0,0 +1,320 @@
...
+        computer_id: str | None = None,
+    ) -> None:
+        if discover_service and self._is_askui_core_service_running():
+            service_msg = (
+                f"Detected running {self._ASKUI_CORE_SERVICE_NAME}; using port "
+                f"{self._ASKUI_CORE_SERVICE_PORT} (controller managed by service)"
+            )
+            logger.info(service_msg)
+            address = _replace_port(address, self._ASKUI_CORE_SERVICE_PORT)
+            is_service = True
```
programminx-askui: Can we move this to an function `discover_service`

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,320 @@
...
+            is_service = True
+        super().__init__(
+            address=address,
+            description=description,
+            display=display,
+            computer_id=computer_id,
+        )
+        self._is_service = is_service
+        self._settings = settings or AskUiControllerSettings()
+        self._process: subprocess.Popen[bytes] | None = None
```
programminx-askui, line 195: If we have a Service then we don't need a `_process`

src/askui/tools/askui/agent_os_server.py
```diff
@@ -0,0 +1,320 @@
...
+        """Whether the server process is managed externally (skip `start()`)."""
+        return self._is_service
+
+    @staticmethod
+    def _is_askui_core_service_running() -> bool:
+        """Return `True` when the `AskuiCoreService` Windows service is RUNNING."""
+        if sys.platform == "win32":
+            try:
+                result = subprocess.run(
+                    ["sc", "query", LocalAgentOsServer._ASKUI_CORE_SERVICE_NAME],
```
programminx-askui: Do we know which rights we need to call `sc` ?

src/askui/tools/askui/agent_os_target_computer.py
programminx-askui: I'm thinking to move this to the askui-agent-os package.

src/askui/tools/askui/agent_os_server.py
```diff
@@ -0,0 +1,320 @@
...
+                )
+            except (OSError, subprocess.SubprocessError):
+                error_msg = (
+                    "Failed to query "
+                    f"{LocalAgentOsServer._ASKUI_CORE_SERVICE_NAME} service"
+                )
+                logger.debug(error_msg)
+                return False
+            if result.returncode != 0:
+                return False
```
programminx-askui: can we move the return block in the try block?

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,320 @@
...
+    def _parse_port(self) -> int:
+        addr = self._address if "://" in self._address else "//" + self._address
+        parsed = urlparse(addr)
+        if parsed.port is None:
+            error_msg = (
+                f"Could not parse port from address {self._address!r}. "
+                "Expected format 'host:port' (e.g. 'localhost:23000')."
+            )
+            raise ValueError(error_msg)
+        return parsed.port
```
programminx-askui, line 265: Is this code not duplicated with __replace_port?

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,320 @@
...
+        commands = [str(path)]
+        if args:
+            commands.extend(args.split())
+        if not logger.isEnabledFor(logging.DEBUG):
+            self._process = subprocess.Popen(
+                commands, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+        else:
+            self._process = subprocess.Popen(commands)
+        wait_for_port(self._parse_port())
```
programminx-askui, line 281: Can we implement to wait for the logs?

src/askui/tools/askui/agent_os_server_manager.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        if server.is_local and any(s.is_local for s in self._servers):
+            existing = next(s for s in self._servers if s.is_local)
+            error_msg = (
+                "Cannot register a second local Agent OS server. At most one local "
+                f"server is supported. Existing local server: "
+                f"{existing.description!r} (computer_id={existing.computer_id!r}). "
+                "Remove it first via `remove(computer_id)`."
+            )
+            raise ValueError(error_msg)
```
programminx-askui: Should this not be handeled by the LocalAgentOS? when he can not aquire a session

src/askui/tools/askui/agent_os_server_manager.py
```diff
@@ -0,0 +1,195 @@
...
+        if any(s.session_guid == server.session_guid for s in self._servers):
+            error_msg = (
+                f"An Agent OS server with session_guid={server.session_guid} is "
+                "already registered. Each server must have a unique session GUID."
+            )
+            raise ValueError(error_msg)
+        if any(s.computer_id == server.computer_id for s in self._servers):
```
programminx-askui: Valid

src/askui/tools/askui/agent_os_server_manager.py
```diff
@@ -0,0 +1,195 @@
...
+        if server.is_local and any(s.is_local for s in self._servers):
+            existing = next(s for s in self._servers if s.is_local)
+            error_msg = (
+                "Cannot register a second local Agent OS server. At most one local "
+                f"server is supported. Existing local server: "
+                f"{existing.description!r} (computer_id={existing.computer_id!r}). "
+                "Remove it first via `remove(computer_id)`."
+            )
+            raise ValueError(error_msg)
+        if any(s.session_guid == server.session_guid for s in self._servers):
```
programminx-askui: Why?

programminx-askui: This should be handeled after connecting.

src/askui/tools/askui/agent_os_server_manager.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Intended for internal lookups (e.g. mapping a gRPC session GUID back to
+        its server during teardown). For user-facing selection, prefer `get`.
+        """
+        for server in self._servers:
+            if server.session_guid == session_guid:
+                return server
+        return None
+
+    def switch(self, computer_id: str) -> AgentOsServer:
```
programminx-askui: We can switch to the same computer? Should this not failing?

src/askui/tools/askui/agent_os_server_manager.py
```diff
@@ -0,0 +1,195 @@
...
+        Args:
+            computer_id (str): The computer id of the server to activate.
+
+        Returns:
+            AgentOsServer: The newly active server.
+
+        Raises:
+            KeyError: If no server with the given computer id is registered.
+        """
+        server = self.get(computer_id)
```
programminx-askui: Please wrap a proper AgentOSServerManager Error around.

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,1296 +1,1543 @@
...
+        self._pre_action_wait = 0
+        self._post_action_wait = 0.05
+        self._max_retries = 10
+        self._reporter = reporter
+        self._manager = AgentOsServerManager(agent_os_servers=agent_os_servers)
+
+    @property
+    def agent_os_server_manager(self) -> AgentOsServerManager:
+        """The underlying Agent-OS-server manager."""
+        return self._manager
```
programminx-askui: Why do we have the server_manager inside the client?

programminx-askui: THis should not be here

src/askui/tools/computer/get_mouse_position_tool.py
```diff
@@ -8,12 +8,20 @@ class ComputerGetMousePositionTool(ComputerBaseTool):
...
+                "server. The result is prefixed with the active Agent OS server "
+                "session GUID."
+            ),
             agent_os=agent_os,
             required_tags=[ToolTags.SCALED_AGENT_OS.value],
         )
         self.is_cacheable = True
 
     def __call__(self) -> str:
+        server = self.agent_os.get_active_agent_os_server(report=False)
```
programminx-askui: Intresting.

tests/unit/tools/askui/test_agent_os_target_computer.py
```diff
@@ -0,0 +1,140 @@
...
+        assert re.fullmatch(
+            r"\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\}",
+            guid,
+        )
+
+    def test_each_generated_guid_is_unique(self) -> None:
+        assert _generate_session_guid() != _generate_session_guid()
+
+
+class TestReplacePort:
```
programminx-askui, line 12: Nice

programminx-askui (changes requested):

src/askui/tools/android/ppadb_agent_os.py
mlikasam-askui: temporary_select is not intended to be called directly by the agent. It is designed to allow users to execute commands on a specific device without needing to restore the previous device context afterward.

It can also be used in custom tools to restrict a tool to a specific device. For example, an enableWifiOnCarEmulator tool could internally use temporary_select, ensuring it only operates on the car emulator. This differs from a generic enableWifi tool, which works across all devices, where the agent is responsible for switching to the appropriate device before invoking the tool.

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -14,37 +14,26 @@
...
-    connect to.
+    Base class describing an Agent OS target computer that the
+    `AskUiControllerClient` can connect to.
 
-    An Agent OS server is the server-side counterpart of the `AgentOs` client
-    abstraction. It runs on the machine being automated, exposes a gRPC API for
-    OS-level operations (screenshot, mouse, keyboard, ...), and is identified by
-    a unique session GUID. Each server also tracks which display it is currently
-    operating against.
+    A target computer runs the server-side counterpart of the `AgentOs` client
```
programminx-askui: AgentOS

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -180,24 +163,47 @@ def _is_askui_core_service_running() -> bool:
         if sys.platform == "win32":
             try:
                 result = subprocess.run(
-                    ["sc", "query", LocalAgentOsServer._ASKUI_CORE_SERVICE_NAME],
+                    [
+                        "sc",
+                        "query",
+                        LocalAgentOsTargetComputer._ASKUI_CORE_SERVICE_NAME,
```
programminx-askui: Which user rights do you need to perfrom the query? Did you tested it without admin rights?

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+            ValueError: If another local target is already registered, the same
+                session GUID or computer id is already registered, or another
+                remote target with the same address is already registered.
+        """
+        self._validate_addable(target)
+        self._by_computer_id[target.computer_id] = target
+        if self._active_computer_id is None:
+            self._active_computer_id = target.computer_id
+        if self.is_connected:
+            self.connect(target)
```
programminx-askui: ~~Why? When you have already on connected target in the list and then you add another unconnected new target. Then it does not connect the new target.~~

Why do we automaticaly connect to the target? This is only an `add`-function.  Then I would prefere a param `auto_connect = True`. And replace the buggy implementation, when one target is connected, that the next one get's not connected.

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+        Args:
+            address (str): gRPC address of the remote Agent OS target computer.
+            description (str): Human-readable description.
+
+        Returns:
+            RemoteAgentOsTargetComputer: The newly registered target.
+        """
+        target = RemoteAgentOsTargetComputer(address=address, description=description)
+        self.add(target)
+        return target
```
programminx-askui: Can we delete this function? It does not make sense.

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+        Returns:
+            AgentOsTargetComputer: The newly active target.
+
+        Raises:
+            KeyError: If no target with the given computer id is registered.
+        """
+        target = self._require(computer_id)
+        self._active_computer_id = computer_id
+        if self.is_connected and computer_id not in self._connections:
+            self.connect(target)
```
programminx-askui: Here you trying to fix the logic from above.

programminx-askui: What should be the Connection Livecycle of an Target? Is the Manager Responsible to connect to the targets?

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+                "`agent_os_target_computers` argument, or call "
+                "`add_agent_os_target_computer()` / "
+                "`add_remote_agent_os_target_computer()` before `connect()`."
+            )
+            raise AskUiControllerError(error_msg)
+        try:
+            for target in self._by_computer_id.values():
+                self.connect(target)
+        except Exception:
+            self.disconnect_all()
```
programminx-askui: Whats happend when an exception is happing on the disconnect function?

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+            if hasattr(e, "add_note"):
+                e.add_note(
+                    f"While connecting to Agent OS target computer "
+                    f"{target.description!r} "
+                    f"(computer_id={target.computer_id!r}, "
+                    f"session_guid={target.session_guid}, "
+                    f"display={target.display}, "
+                    f"address={target.address})"
+                )
+            raise
```
programminx-askui: Why aer you not creating a new Exception? You should not manipulate a old one.

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+            started_process = True
+        channel = grpc.insecure_channel(
+            target.address,
+            options=[
+                ("grpc.max_send_message_length", 2**30),
+                ("grpc.max_receive_message_length", 2**30),
+                ("grpc.default_deadline", 300000),
+            ],
+        )
+        stub = controller_v1.ControllerAPIStub(channel)
```
programminx-askui: Why do we have inside the Manager GRPC Code? This should be moved to the Client. We need to get rid of this.

src/askui/tools/askui/askui_controller.py
```diff
@@ -80,232 +78,216 @@
...
-class _Connection:
-    """gRPC connection state for a single Agent OS server."""
-
-    channel: grpc.Channel
-    stub: controller_v1.ControllerAPIStub
-    session_info: controller_v1_pbs.SessionInfo
-    started_process: bool
-
 
 class AskUiControllerClient(AgentOs):
```
programminx-askui: Why does a Client has a Manager with multiple Targets? This should be another way around: Manager has Multiple Target and the Target should have the Client.

programminx-askui (changes requested):
Let's have a talk.

src/askui/tools/askui/askui_controller.py
programminx-askui: Rename to: `MultiComputerTargetAgentOS`

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+    """gRPC connection state for a single Agent OS target computer."""
+
+    target: AgentOsTargetComputer
+    channel: grpc.Channel
+    stub: controller_v1.ControllerAPIStub
+    session_info: controller_v1_pbs.SessionInfo
+    started_process: bool
+
+
+class AgentOsTargetComputerManager:
```
programminx-askui: Rename to `ComputerTargetPool``

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,329 @@
...
+
+from typing_extensions import override
+
+from askui.tools.askui.askui_controller_settings import AskUiControllerSettings
+from askui.tools.utils import process_exists, wait_for_port
+
+logger = logging.getLogger(__name__)
+
+
+class AgentOsTargetComputer:
```
programminx-askui:
```suggestion
class ComputerTarget
```

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,329 @@
...
+    def __repr__(self) -> str:
+        return (
+            f"{type(self).__name__}("
+            f"computer_id={self._computer_id!r}, "
+            f"description={self._description!r}, "
+            f"display={self._display!r})"
+        )
+
+
+class LocalAgentOsTargetComputer(AgentOsTargetComputer):
```
programminx-askui:
```suggestion
class LocalComputerTarget(ComputerTarget):
```

src/askui/tools/askui/agent_os_target_computer.py
```diff
@@ -0,0 +1,329 @@
...
+                    self.clean_up()
+            else:
+                self._process.terminate()
+        except Exception:  # noqa: BLE001 - We want to catch all other exceptions here
+            logger.exception("Error stopping local controller process")
+        finally:
+            self._process = None
+
+
+class RemoteAgentOsTargetComputer(AgentOsTargetComputer):
```
programminx-askui:
```suggestion
class RemoteComputerTarget(ComputerTarget):
```

src/askui/tools/askui/__init__.py
```diff
@@ -1,6 +1,17 @@
-from .askui_controller import AskUiControllerClient, AskUiControllerServer
+from .agent_os_target_computer import (
```
programminx-askui, line 1: can we rename the folder from `askui` -> `agent_os` so it's clear that  it's the AskUI AgentOS and not askui related stuff

src/askui/tools/askui/agent_os_target_computer_manager.py
programminx-askui: I still see no benefit of this funciton. -> please remove

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+        """
+        target = RemoteAgentOsTargetComputer(address=address, description=description)
+        self.add(target)
+        return target
+
+    def reset(self) -> None:
+        """Disconnect every open connection and remove all registered targets."""
+        self.disconnect_all()
+        self._by_computer_id.clear()
+        self._active_computer_id = None
```
programminx-askui: Whats happend when an exception happens during the disconnct. then the computer_id are not cleared


```suggestion
        try:
            self.disconnect_all()
        finally:
           self._by_computer_id.clear()
            self._active_computer_id = None
```

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+
+        Args:
+            computer_id (str): The computer id of the target to remove.
+
+        Raises:
+            KeyError: If no target with the given computer id is registered.
+        """
+        self._require(computer_id)
+        self.disconnect(computer_id)
+        del self._by_computer_id[computer_id]
```
programminx-askui: are you removing with this the dict entry too?

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+        if self._active_computer_id == computer_id:
+            self._active_computer_id = next(iter(self._by_computer_id), None)
+
+    def list(self) -> list[AgentOsTargetComputer]:
+        """Return all registered targets in registration order."""
+        return list(self._by_computer_id.values())
```
programminx-askui: why to we need a list function? 

You are exposing internals to other Methods. This is an anti pattern. A consumer of the list function could now disconnect the entires. 

As far as I saw it's only used as a `__str__` or `__repre__` funciton

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+
+    def get(self, computer_id: str) -> AgentOsTargetComputer:
```
programminx-askui: Same as the `list` functions. I'm here a little bit unhappy.

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+                "No active Agent OS target computer. Register one via "
+                "`AskUiControllerClient.add_agent_os_target_computer()` / "
+                "`add_remote_agent_os_target_computer()`, or pass "
+                "`agent_os_target_computers` to the `AskUiControllerClient` "
+                "constructor."
+            )
+            raise AskUiControllerError(error_msg)
+        return target
+
+    def active_connection(self) -> _Connection:
```
programminx-askui: Why is the Connection not part of the ComputerTarget? And we have to manage this in an extra list?

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+            error_msg = (
+                f"Active Agent OS target computer {target.description!r} "
+                f"(computer_id={target.computer_id!r}, "
+                f"address={target.address}) "
+                "is not connected. Call `AskUiControllerClient.connect()` first."
+            )
+            raise AskUiControllerError(error_msg)
+        return conn
+
+    def connect_all(self) -> None:
```
programminx-askui: The Pool implies already that we connect to all Target. So we do not have to duplicate it. 

```suggestion
    def connect(self) -> None:
```

src/askui/tools/askui/agent_os_target_computer_manager.py
```diff
@@ -0,0 +1,414 @@
...
+            raise
+        self._connections[target.computer_id] = _Connection(
+            target=target,
+            channel=channel,
+            stub=stub,
+            session_info=session_info,
+            started_process=started_process,
+        )
+
+    def disconnect_all(self) -> None:
```
programminx-askui: Same here

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,1415 +1,1533 @@
...
+    def _require_active_agent_os_target_computer(self) -> AgentOsTargetComputer:
+        return self._manager.require_active()
+
+    @property
+    def _session_info(self) -> controller_v1_pbs.SessionInfo:
+        return self._manager.active_connection().session_info
+
+    @telemetry.record_call()
+    @override
+    def add_remote_agent_os_target_computer(
```
programminx-askui: Can we not have a `add` function with accepts

```suggestion
    def add(computer_target: RemoteComputerTargetOptions | LocalComputerTargetOptions
```

src/askui/tools/askui/askui_controller.py
```diff
@@ -1,1415 +1,1533 @@
...
+        )
+        self._manager.add(agent_os_target_computer)
+        return agent_os_target_computer
+
+    @telemetry.record_call(exclude={"agent_os_target_computers"})
+    @override
+    def reset_agent_os_target_computers(
+        self,
+        agent_os_target_computers: list[AgentOsTargetComputer] | None = None,
+    ) -> None:
```
programminx-askui, line 175: This function should not accept a list. It should use the current pool and reset it

```suggestion
    def reset_computer_targets(
        self,
    ) -> None:
```

src/askui/tools/agent_os.py
```diff
@@ -677,6 +683,65 @@ def set_window_in_focus(self, process_id: int, window_id: int) -> None:
...
+        Example:
+            ```python
+            with agent_os.temporary_select('Remote-Machine') as remote_machine:
+                img = remote_machine.screenshot()
+                img.save("remote_machine.png")
+            # previous active target restored here
+            ```
+        """
+        raise NotImplementedError
+
```
programminx-askui, line 738: This should not be part of the default AgentOS 

-> Please Remove

src/askui/tools/computer/switch_agent_os_target_computer_tool.py
```diff
@@ -0,0 +1,28 @@
+from askui.models.shared import ComputerBaseTool
+from askui.tools.agent_os import AgentOs
+
+
+class ComputerSwitchAgentOsTargetComputerTool(ComputerBaseTool):
+    def __init__(self, agent_os: AgentOs | None = None) -> None:
```
programminx-askui: The init should either accept only MultiComputerTargetAgentOS or check if it's an instance of,

src/askui/tools/agent_os.py
programminx-askui: -> This should be part of an ComputerAgentOS

src/askui/tools/computer/switch_agent_os_target_computer_tool.py
programminx-askui: So all ComputerBaseTools should only accept an `ComputerAgentOS`

programminx-askui (changes requested):
Hello @mlikasam-askui , 

I added some changes. But we have to do after this more work.

src/askui/tools/askui/computer_target_connection.py
```diff
@@ -0,0 +1,131 @@
...
+        try:
+            self.channel.close()
+        except Exception:  # noqa: BLE001
+            logger.exception("Error closing channel for controller %s", computer_id)
+        try:
+            target.stop()
+        except Exception:  # noqa: BLE001
+            logger.exception(
+                "Error stopping client-started controller process for %s", computer_id
+            )
```
programminx-askui, line 128: This exception handling will shoot us in the future in the food. But let's keep it for now

src/askui/tools/askui/computer_target_pool.py
```diff
@@ -0,0 +1,281 @@
...
+        target.connect()
+
+    def disconnect(self) -> None:
+        """
+        Close every open Agent OS target connection. Errors on one connection
+        are logged but do not abort the loop - a partial failure still releases
+        the others.
+        """
+        for target in self._by_computer_id.values():
+            target.disconnect()
```
programminx-askui, line 217: What's happend when one of the disconnect failed? SO the others does not get disconnected?

programminx-askui (approved):
Let's merge it in. And check where it crashes. :-D 

@philipph-askui we need to be careful to find the breaking changes.

philipph-askui:
The following is still missing for me:
1) updated docs
2) an example
3) How can I add a tool that the agent can only execute on one target, e.g. a FileReadTool from a specific base directory that only exists on one target?

PR #277, opened 2026-05-13
feat: add GetFile, GetFileList, ClearVirtualDisplays agent OS commands 

mlikasam-askui:
**Jira:** SOLENG-363, SOLENG-362


src/askui/tools/askui/askui_ui_controller_grpc/generated/AgentOS_Send_Request_2501.py
```diff
@@ -393,6 +394,20 @@ class LoadCharacterMapCommand(BaseModel):
...
+
+
+class GetFileCommand(BaseModel):
+    name: Literal['GetFile'] = 'GetFile'
+    parameters: list[str] = Field(..., max_length=1, min_length=1)
+
+
+class RemoveVirtualDisplaysCommand(BaseModel):
+    name: Literal['RemoveVirtualDisplays'] = 'RemoveVirtualDisplays'
+    parameters: List[Any] = []
```
philipph-askui: can we specify the type here instead of using Any?

src/askui/tools/askui/askui_controller.py
```diff
@@ -217,6 +224,8 @@ def connect(self) -> None:
         self._start_session()
         self._start_execution()
         self.set_display(self._display)
+        if self._settings.clean_virtual_displays:
```
philipph-askui, line 227: shall we add an INFO log message here?

src/askui/tools/askui/askui_controller.py
```diff
@@ -1294,3 +1303,117 @@ def set_window_in_focus(self, process_id: int, window_id: int) -> None:
...
+            (
+                "Setting display to 1 to ensure that the "
+                "controller is using the real display"
+            )
+        )
+        self.set_display(1)
+
+    @staticmethod
+    def _decode_file_payload(base64_data: str) -> Image.Image | str:
+        try:
```
philipph-askui, line 1404: on one hand this omits the need for a "magic" set of endings that we convert, on the other this might be inefficient. Not sure how much time/compute this consumes for huge text files that are clearly not images

src/askui/tools/askui/askui_controller.py
```diff
@@ -1294,3 +1303,117 @@ def set_window_in_focus(self, process_id: int, window_id: int) -> None:
...
+            raise DesktopAgentOsError(res.error)
+        if res.response is None:
+            message = f"{type(res).__name__} is missing both error and response"
+            raise DesktopAgentOsError(message)
+        decoded = self._decode_file_payload(res.response.file.content)
+        if isinstance(decoded, Image.Image):
+            detail = f"image ({decoded.format}, {decoded.size[0]}x{decoded.size[1]})"
+            self._reporter.add_message(
+                "AgentOS", f"get_file({path}) -> {detail}", decoded
+            )
```
philipph-askui, line 1382: why not add the PIL image here as well?

src/askui/tools/askui/askui_controller.py
```diff
@@ -1294,3 +1303,117 @@ def set_window_in_focus(self, process_id: int, window_id: int) -> None:
...
+        self._send_command(command)
+        self._reporter.add_message("AgentOS", "remove_virtual_displays() -> done")
+
+        logger.debug(
+            (
+                "Setting display to 1 to ensure that the "
+                "controller is using the real display"
+            )
+        )
+        self.set_display(1)
```
philipph-askui: will this become a problem if I in the act command I set removing displays to true, but set the display id to 2?

src/askui/tools/store/computer/experimental/get_file.py
```diff
@@ -0,0 +1,54 @@
...
+                            "example 'C:\\\\path\\\\notes.txt' on Windows or "
+                            "'/home/user/notes.txt' on Linux/macOS)."
+                        ),
+                    },
+                },
+                "required": ["absolute_file_path"],
+            },
+            agent_os=agent_os,
+            required_tags=[ToolTags.SCALED_AGENT_OS.value],
+        )
```
philipph-askui, line 51: please add the self.is_cacheable=True flag to init

src/askui/tools/store/computer/experimental/remove_virtual_displays.py
```diff
@@ -0,0 +1,47 @@
...
+                "continuing automation."
+            ),
+            input_schema={
+                "type": "object",
+                "properties": {},
+                "required": [],
+            },
+            agent_os=agent_os,
+        )
+
```
philipph-askui, line 39: please add the self.is_cacheable=True flag to init

philipph-askui (changes requested):

philipph-askui:
What happens if I have the current controller on mac and the agent tries to call the read_file, I assume it will crash?
Can we add a warning to the 2 new tools that they are currently only available under windows and that they will cause the execution to crash under mac/linux?

should we need to update the docs and add a hint, that e.g. the read_file operation is only available on windows currently?

mlikasam-askui:
> What happens if I have the current controller on mac and the agent tries to call the read_file, I assume it will crash? Can we add a warning to the 2 new tools that they are currently only available under windows and that they will cause the execution to crash under mac/linux?
> 
> should we need to update the docs and add a hint, that e.g. the read_file operation is only available on windows currently?

It won’t crash. The agent will receive an error indicating that the currently used control does not support the requested command.

src/askui/tools/askui/askui_controller.py
mlikasam-askui: the `decoded ` response is the PIL image.

src/askui/tools/askui/askui_controller.py
mlikasam-askui: I have removed it.

philipph-askui (approved):

PR #280, opened 2026-05-29
feat: refactors truncation strategy to always keep the first user message

philipph-askui:
the first user message usually contains a lot of critical information. To make sure it is not lost, we should always keep it, even when truncating

Also, this PR updates the truncation trigger threshold from 70k tokens to 112k tokens, increasing the headroom for new messages after truncation by roughly 2x

### DIFF
--- a/src/askui/models/shared/truncation_strategies.py
+++ b/src/askui/models/shared/truncation_strategies.py
@@ -222,6 +222,7 @@ class TruncationStrategy(ABC):
     ) -> None:
         self._full_message_history: list[MessageParam] = []
         self._truncated_message_history: list[MessageParam] = []
+        self._first_user_message: MessageParam | None = None
         self._max_messages = max_messages
         self._absolute_truncation_threshold = int(
             max_input_tokens * truncation_threshold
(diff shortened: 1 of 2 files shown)


src/askui/models/shared/truncation_strategies.py
```diff
@@ -30,10 +30,10 @@
 logger = logging.getLogger(__name__)
 
 # needs to be below limits imposed by endpoint
-MAX_INPUT_TOKENS = 100_000
+MAX_INPUT_TOKENS = 200_000
 
 # we will truncate as soon as we reach this threshold
-TRUNCATION_THRESHOLD = 0.7
+TRUNCATION_THRESHOLD = 0.56
```
mlikasam-askui, line 36: What was the reason for decreasing this threshold?

mlikasam-askui (approved):

src/askui/models/shared/truncation_strategies.py
philipph-askui: I also changed the "MAX_INPUT_TOKENS" value from 100k to 200k to align with the actual limit of the endpoint. 
The only relevant value in the code is the MAX_INPUT_TOKENS*TRUNCATION_THRESHOLD
If we would have kept the TRUNCATION_THRESHOLD value as it was, this value would have become too large (the higher it is, the more expensive conversations can get)

src/askui/models/shared/truncation_strategies.py
```diff
@@ -274,9 +329,13 @@ def reset(self, messages: list[MessageParam] | None = None) -> None:
         if messages is not None:
             self._full_message_history = list(messages)
             self._truncated_message_history = list(messages)
+            self._first_user_message = next(
+                (m for m in messages if m.role == "user"), None
+            )
```
mlikasam-askui: this can be moved to the  `_capture_first_user_message` function

philipph-askui: addressed in 48c1020865776f414b629dd48c9622d3b67e840e

PR #281, opened 2026-06-02
feat: keep requests under the endpoint byte limit by stripping oldest images

mlikasam-askui:
Strip oldest base64 images from history when a request would exceed the endpoint's ~30 MB byte limit. Adds request-size estimation, a per-provider `max_request_bytes`, and budget enforcement in `SummarizingTruncationStrategy` (skipped when no limit is defined).

**Jira:** SOLENG-369

### DIFF
--- a/src/askui/models/anthropic/messages_api.py
+++ b/src/askui/models/anthropic/messages_api.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Any, Tuple, cast
 
 from anthropic import (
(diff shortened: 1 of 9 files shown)


src/askui/models/anthropic/messages_api.py
```diff
@@ -193,6 +200,16 @@ def create_message(
         temperature: float | None = None,
         provider_options: dict[str, Any] | None = None,
     ) -> MessageParam:
+        estimated_bytes = estimate_messages_bytes(messages)
+        if estimated_bytes > ANTHROPIC_MAX_REQUEST_BYTES:
```
programminx-askui, line 204: I don't like that we estimate the Request side. I would prefere to deal with the Error. And then do the Truncation strategy.

Otherwise we don't know if the LLM Provider is inceasing theire 30 MB Limit

programminx-askui (changes requested):

PR #282, opened 2026-06-09
feat: add coordinate space abstraction for open weights LLM support

philipph-askui:
Introduces VlmCoordinateSpace strategy (pixel, scaled, normalized) so agentOS facades can map model-emitted coordinates to screen pixels.
Thereby adds auto-detection for Qwen, Holo (0-1000 grid) and Kimi (0.0-1.0 floats) in OllamaVlmProvider.
Appends coordinate info to system prompts for OpenAI-compatible providers.


src/askui/model_providers/anthropic_vlm_provider.py
```diff
@@ -70,6 +78,8 @@ def __init__(
         auth_token: str | None = None,
         model_id: str | None = None,
         client: Anthropic | None = None,
+        image_scaler: ImageScaler | None = None,
+        max_image_edge: int | None = None,
```
programminx-askui: Everything what is images related should start with `image_`. Similare like `model_`. Otherwise it's hard to understand the zusammenhänge


```suggestion
        image_scaler: ImageScaler | None = None,
        image_edge_max: int | None = None,
```

programminx-askui: but is this not a property of the image scaler?

src/askui/model_providers/anthropic_vlm_provider.py
```diff
@@ -104,6 +120,14 @@ def model_id(self) -> str:
     def pricing(self) -> ModelPricing | None:
         return self._pricing
 
+    @property
+    @override
+    def image_scaler(self) -> ImageScaler:
+        if self._image_scaler_override is not None:
+            return self._image_scaler_override
+        max_edge = self._max_edge
+        return lambda image: compute_patch_optimized_image(image, max_edge=max_edge)
```
programminx-askui: are we sure, that we can use the lambda? I had bad experience with lambda in python, that they caach the computation.  But can't remember

src/askui/model_providers/askui_vlm_provider.py
```diff
@@ -57,18 +63,34 @@ def __init__(
...
     def model_id(self) -> str:
         return self._model_id_value
 
+    @property
+    @override
+    def image_scaler(self) -> ImageScaler:
+        if self._image_scaler_override is not None:
+            return self._image_scaler_override
+        max_edge = self._max_edge
+        return lambda image: compute_patch_optimized_image(image, max_edge=max_edge)
```
programminx-askui: Why so complicated?


```suggestion
        self._image_scaler = image_scaler if image_scaler else lambda image: compute_patch_optimized_image(image, max_edge=max_edge)
        self._max_edge = (
            max_image_edge
            or int(os.environ.get("ASKUI_VLM_MAX_IMAGE_EDGE", "0"))
            or _DEFAULT_MAX_IMAGE_EDGE
        )

@property
@override
def model_id(self) -> str:
return self._model_id_value

```

src/askui/model_providers/askui_vlm_provider.py
```diff
@@ -57,18 +63,34 @@ def __init__(
...
         self._model_id_value = (
             model_id or os.environ.get("VLM_PROVIDER_MODEL_ID") or _DEFAULT_MODEL_ID
         )
         self._injected_client = client
+        self._image_scaler_override = image_scaler
+        self._max_edge = (
+            max_image_edge
+            or int(os.environ.get("ASKUI_VLM_MAX_IMAGE_EDGE", "0"))
+            or _DEFAULT_MAX_IMAGE_EDGE
+        )
```
programminx-askui: I would remove it if possible

src/askui/tools/playwright/agent_os_facade.py
```diff
@@ -2,65 +2,59 @@
...
+        )
         self.tags = self._agent_os.tags + [ToolTags.SCALED_AGENT_OS.value]
 
     def connect(self) -> None:
         self._agent_os.connect()
-        self._real_screen_resolution = self._agent_os.screenshot(
-            report=False,
+        self._scaler.real_screen_resolution = self._agent_os.screenshot(
+            report=False
         ).size
```
programminx-askui, line 52: The screen resolution can change over time

src/askui/tools/playwright/agent_os_facade.py
```diff
@@ -2,65 +2,59 @@
...
+        coordinate_space: VlmCoordinateSpace,
+        image_scaler: ImageScaler,
+    ) -> None:
         self._agent_os = agent_os
-        self._target_resolution: tuple[int, int] = (1024, 768)
-        self._real_screen_resolution: tuple[int, int] | None = None
+        self._scaler = CoordinateScaler(
+            coordinate_space=coordinate_space,
+            image_scaler=image_scaler,
+            fetch_real_resolution=lambda: self._agent_os.screenshot(report=False).size,
```
programminx-askui: I'm not a fan of lambda in python. What is the reason to use them?

src/askui/tools/android/agent_os_facade.py
```diff
@@ -1,74 +1,76 @@
...
+    def __init__(
+        self,
+        agent_os: AndroidAgentOs,
+        coordinate_space: VlmCoordinateSpace,
+        image_scaler: ImageScaler,
+    ) -> None:
         self._agent_os: AndroidAgentOs = agent_os
-        self._target_resolution: Tuple[int, int] = (1024, 768)
-        self._real_screen_resolution: Optional[Tuple[int, int]] = None
+        self._scaler = CoordinateScaler(
```
programminx-askui, line 38: This is a note.

The "CoordinateScaler" should be on the DisplayCapability/DisplayTarget/WindowTarget/WindowCapability.

But here we assume, that we have only one ClickableTarget.

src/askui/tools/coordinate_scaler.py
```diff
@@ -0,0 +1,99 @@
...
+        self.real_screen_resolution: tuple[int, int] | None = None
+
+    def scale_screenshot(self, screenshot: Image.Image) -> Image.Image:
+        """Record real resolution, apply scaler, record target resolution."""
+        self.real_screen_resolution = screenshot.size
+        scaled = self._image_scaler(screenshot)
+        self.target_resolution = scaled.size
+        return scaled
+
+    def scale_coordinates(
```
programminx-askui, line 53: Do we need this function? Should this not replaced from the ScaleCoordinate Layer?

src/askui/utils/llm_image_utils.py
```diff
@@ -0,0 +1,191 @@
...
+    # Binary search for largest scale that fits within token budget
+    lo, hi = 0.0, scale
+    for _ in range(50):
+        mid = (lo + hi) / 2
+        w = max(1, int(width * mid))
+        h = max(1, int(height * mid))
+        if count_image_tokens(w, h, patch_size) <= max_tokens:
+            lo = mid
+        else:
+            hi = mid
```
programminx-askui, line 80: do we need this? can we move this to a own function?

tests/unit/tools/test_agent_os_facade_coordinates.py
```diff
@@ -0,0 +1,160 @@
...
+
+import pytest
+from PIL import Image
+
+from askui.models.shared.coordinate_space import (
+    NormalizedCoordinateSpace,
+    PixelCoordinateSpace,
+    ScaledCoordinateSpace,
+)
+from askui.tools.android.agent_os_facade import AndroidAgentOsFacade
```
programminx-askui, line 18: Did you checked the tests? I would add here some dynamic tests. with different resolutions and negative tests

programminx-askui (commented):
Hello, 

I've check the PR. Overall nice. We should introduce a ClickableCapability/ClickableTarget.

src/askui/model_providers/anthropic_vlm_provider.py
philipph-askui: yes. This value is only used if users do not specify an ImageScaler explicitly but the default one is used.

src/askui/model_providers/anthropic_vlm_provider.py
philipph-askui: I refactored this to that we now have a proper ImageScaler base class with subclasses that are used here instead.

src/askui/model_providers/askui_vlm_provider.py
philipph-askui: you mean the env variable?

src/askui/utils/llm_image_utils.py
philipph-askui: code was copied directly from an example from anthropic, so I would keep it here as is

tests/unit/tools/test_agent_os_facade_coordinates.py
philipph-askui: done

src/askui/tools/playwright/agent_os_facade.py
philipph-askui: this value will be update with every new screenshot

src/askui/model_providers/askui_vlm_provider.py
philipph-askui: I refactored it to use a proper imagescaler class i/o lambdas

src/askui/tools/playwright/agent_os_facade.py
philipph-askui: removed the lambdas

src/askui/tools/android/agent_os_facade.py
philipph-askui: at the moment it does not matter after all what target we are scaling for. The llm sees a screenshot and predicts a position on it. All of what we are doing is scaling the predicted coordinates to the correct absolute image coordinates of the screenshot we got

src/askui/tools/coordinate_scaler.py
philipph-askui: this is the scaleCoordinate Layer :-D or what do you mean?

PR #283, opened 2026-06-10
Feat/tool result image scaling

philipph-askui:
Background:
anthropic imposes a maximum image size of 2000x2000px for images if we have more than 20 images in the message history.

We need to discuss if we want to handle downwscaling here for all tool result images that are too large or if we want to have this as a tool responsibility!

Either way, we need to add a method to image utils that can be used by tools or our sdk to downscale the image outputs.

### DIFF
--- a/docs/07_tools.md
+++ b/docs/07_tools.md
@@ -181,6 +181,18 @@ A tool’s __call__ method may return:
 - None
 - a list or tuple containing any of the above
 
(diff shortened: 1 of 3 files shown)


programminx-askui (approved):

mlikasam-askui (approved):

philipph-askui:
I added this directly to PR #282 (see #fabd05d0bae01539604e2539fe9cfb9386f7846f) and hence will close this PR

PR #289, opened 2026-06-26
feat: support PDF documents in tools, get(), and file reads

mlikasam-askui:
## Summary

Adds first-class PDF support across the SDK, mirroring how images are already
handled. PDFs are sent to the model as a `document` content block — an
Anthropic base64 `document` block (no beta header) and an OpenAI `file`
content part.

## Changes

- **Message types**: new `DocumentBlockParam` + `Base64PdfSourceParam`; allowed
  in `ToolResultBlockParam` content and exported from `askui` / `askui.models`.
- **Tool results**: a tool that returns a `PdfSource` is rendered as a document
  block (`_convert_to_content`). MCP tools returning a PDF as an embedded blob
  resource (`application/pdf`) are now converted too, instead of being dropped.
- **`get()`**: the Anthropic and OpenAI get models accept `PdfSource` (Office
  documents remain unsupported).
- **File reads**: `AgentOs.get_file` detects and returns PDFs as `PdfSource`;
  file-type detection now sniffs the MIME via `filetype.guess` and decodes the
  base64 payload once.
- **`LoadPdfTool`**: new universal tool to load a PDF from disk and hand it to
  the model (the PDF counterpart to `LoadImageTool`).
- **Reporting**: base64 PDF (and any media) blobs are truncated in reports to
  keep them readable.
- **Chore**: exclude `venv` from the mypy typecheck.


philipph-askui:
Claude review:

1. 🟠 PDF document blocks are never truncated in long act() loops — truncation_strategies.py:603

The sliding-window strategy's new_nested type was widened to include DocumentBlockParam, but no stripping logic was added — documents fall through the else: new_nested.append(nested) branch untouched. Only ImageBlockParam is stripped. A PDF is a far larger base64 blob than a screenshot, so in a multi-turn act() run a tool result containing a PDF stays in context forever, while screenshots get pruned. This is the highest-impact finding: real token/cost/context-window growth, and the widened annotation makes it look handled when it isn't. Either strip stale document blocks the same way, or document explicitly why PDFs are exempt.

2. 🟠 No size/page guard before sending — built_messages_for_get_pdf, LoadPdfTool, get_file

The entire PDF is base64-encoded and sent with no client-side check. Anthropic enforces 32 MB / 100 pages per request; OpenAI has its own caps. A user pointing get("...", source="big.pdf") or LoadPdfTool at a 200-page report gets an opaque 400/413 from the provider instead of a clear SDK-side error. Given the SDK already has friendly exceptions, a pre-flight size/page check (or at least a documented limit) would be a meaningful UX improvement.

3. 🟡 Two MCP conversion paths still don't know about PDFs — asymmetry in tools.py

The PR taught the primary inbound path (_convert_call_tool_result, CallToolResult "resource" blocks) about PDFs, but two sibling paths were left behind:
- _convert_to_mcp_content (tools.py:159, outbound): handles tuple/list/Image.Image but not PdfSource — returns it raw. Any askui tool returning a PDF that is exposed as an MCP server (to_mcp_tool, used by the Chat API) will hand FastMCP a raw PdfSource pydantic model, which won't serialize to a blob resource correctly.
- _convert_from_mcp_tool_call_result (tools.py:175, the _McpToolAdapter path): only handles McpTextContent/McpImageContent and raises McpToolAdapterException on a PDF resource. So an external MCP tool returning a PDF works through the act-loop CallToolResult path but throws through Tool.from_mcp_tool.

The result is that "PDFs round-trip through MCP" is true on one path and false on two others. Worth either closing the gap or documenting the limitation.

4. 🟡 Base64PdfSourceParam.data is unvalidated, PdfSource trusts its input

PdfSource(path) / PdfSource(bytes) perform no PDF sniffing — only get_file validates via filetype.guess. LoadPdfTool will happily wrap a misnamed .pdf that isn't one and ship it to the model as application/pdf. Low severity (the model degrades gracefully), but a magic-bytes check in LoadPdfTool.__call__ would catch it early and match the rigor of the get_file path.

5. 🔵 Nits

- Hardcoded "document.pdf" filename in three places (openai/get_model.py, openai/messages_api.py). For LoadPdfTool/get_file the real filename is known and would help the model; consider threading it through.
- cache_control on DocumentBlockParam is never set anywhere. PDFs are the most expensive block to re-send each turn — they're the prime candidate for cache_control: ephemeral. (Combined with finding #1, an un-truncated, un-cached PDF re-sent every turn is the worst case.)
- mypy.ini now excludes ^\.?venv/.*$ — unrelated to PDF support; fine, but it's scope creep riding along in a feature commit.
- LoadPdfTool raises FileExistsError when the path is a directory — semantically odd (that's normally for "already exists" collisions); IsADirectoryError or ValueError reads better.

PR #291, opened 2026-07-01
feat: add scratch_act.py to demonstrate ComputerAgent functionality

danyalxahid-askui:
It updates the `thinking` configuration in `ComputerAgent` to use an adaptive strategy. Should be changed for non claude-5 models.

tested with changing env variable "VLM_PROVIDER_MODEL_ID" to "claude-sonnet-5", "gemini-3.5-flash"

### DIFF
--- a/scratch_act.py
+++ b/scratch_act.py
@@ -1,0 +1,4 @@
+from askui import ComputerAgent
+
+with ComputerAgent() as agent:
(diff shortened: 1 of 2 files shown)


philipph-askui:
Closing this PR as we now implemented this in #292

PR #305, opened 2026-08-10
fix: surface Desktop Agent OS errors to the agent instead of crashing

philipph-askui:
## Problem

When trying to read a remote file or list a directory that does not exist the whole agent run crashed with an unhandled error, instead of the failure being reported back to the agent as a tool error (the way every other tool failure is handled).

The reason: the Desktop Agent OS error type inherited from `BaseException` rather than `Exception`. The tool-calling loop catches tool failures with an `except Exception` handler and converts them into an error result the agent can see and react to. Because `except Exception` does not catch `BaseException` subclasses, this error slipped past the handler and propagated all the way up, crashing the run.

By convention, `BaseException` is reserved for system-level signals (like `KeyboardInterrupt`/`SystemExit`); application errors should subclass `Exception`.

## Fix

Make the Desktop Agent OS error type inherit from `Exception`. No other changes are needed — the existing error-handling logic now catches it and returns it to the agent as a tool error result, consistent with how the rest of the codebase handles tool failures.

## Testing

- Added a regression test that drives a tool raising a Desktop Agent OS error through the tool-calling loop and asserts an error result is returned rather than the exception propagating.
- `pdm run qa:fix` passes (typecheck, format, lint).
- New and related unit tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/askui/tools/askui/askui_ui_controller_grpc/desktop_agent_os_error.py
+++ b/src/askui/tools/askui/askui_ui_controller_grpc/desktop_agent_os_error.py
@@ -1,5 +1,9 @@
-class DesktopAgentOsError(BaseException):
+class DesktopAgentOsError(Exception):
     """Base class for Desktop Agent OS errors.
 
     This error is raised when an error occurs in the Desktop Agent OS.
+
+    Inherits from `Exception` (not `BaseException`) so that the standard
+    `except Exception` handlers in the tool-calling loop can catch it and
+    surface it to the agent as a tool error result instead of crashing.
     """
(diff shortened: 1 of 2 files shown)


src/askui/tools/askui/askui_ui_controller_grpc/desktop_agent_os_error.py
```diff
@@ -1,5 +1,9 @@
-class DesktopAgentOsError(BaseException):
+class DesktopAgentOsError(Exception):
```
mlikasam-askui, line 1: I thought we wanted to have two types: DesktopAgentOsError for errors that aren’t fixable, and DesktopAgentOsException for errors that are recoverable by the agent.

mlikasam-askui (approved):

src/askui/tools/askui/askui_ui_controller_grpc/desktop_agent_os_error.py
philipph-askui: You are right. I guess then I'll just close this PR?

PR #308, opened 2026-08-12
fix(models): request visible thinking summaries on adaptive-thinking models

philipph-askui:
## Why

Models of the Sonnet 5 generation onward default the thinking `display` setting to **"omitted"**: the API returns thinking blocks whose text is empty (signature only), while the full thinking tokens are still billed (`usage.output_tokens_details.thinking_tokens`). Every consumer of `make_thinking_settings()` therefore silently loses all visible reasoning in reports and logs on the newest models — bitten in production by the TCPos test-automation project, where HTML reports showed 175/175 empty thinking blocks.

Older adaptive models (Sonnet 4.6 / Opus 4.6) default `display` to summarized, which is why the gap only appears from Sonnet 5 onward.

## What

`make_thinking_settings()` now returns `thinking={"type": "adaptive", "display": "summarized"}` for adaptive models. Billing is identical for both display modes ([docs](https://platform.claude.com/docs/en/build-with-claude/thinking-steering-and-cost#pricing)) — this only restores visibility. The legacy `budget_tokens` branch is unchanged.

## Verification

- `pytest tests/unit/models/test_thinking.py` — 54 passed (assertions updated)
- `ruff check` / `ruff format --check` clean on changed files
- Live end-to-end against Vertex `claude-sonnet-5` (`rawPredict`, eu + global endpoints): empty thinking text before, 3,162 chars of summarized thinking after — with identical `thinking_tokens` billed

## Notes

Same follow-up as #302: the C# SDK carries a port of this module — mirroring PR opened in askui/csharp-sdk.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/askui/models/shared/thinking.py
+++ b/src/askui/models/shared/thinking.py
@@ -159,10 +159,17 @@ def make_thinking_settings(
             **make_thinking_settings(self._vlm_provider.model_id),
         )
 
-    Models that support adaptive thinking get ``thinking={"type": "adaptive"}``
-    (with ``effort`` sent via ``provider_options["output_config"]`` when given);
-    older models get a fixed token budget of
-    ``thinking={"type": "enabled", "budget_tokens": 2048}`` and ignore ``effort``.
+    Models that support adaptive thinking get
+    ``thinking={"type": "adaptive", "display": "summarized"}`` (with ``effort``
+    sent via ``provider_options["output_config"]`` when given); older models
+    get a fixed token budget of
+    ``thinking={"type": "enabled", "budget_tokens": 2048}`` and ignore
+    ``effort``. ``display`` is set explicitly because the newest adaptive
+    models (Sonnet 5 generation onward) default it to ``"omitted"``, which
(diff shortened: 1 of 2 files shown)


src/askui/models/shared/thinking.py
```diff
@@ -176,7 +183,9 @@ def make_thinking_settings(
             when applicable, ``provider_options``).
     """
     if uses_adaptive_thinking(model_id):
-        settings: dict[str, Any] = {"thinking": {"type": "adaptive"}}
+        settings: dict[str, Any] = {
```
mlikasam-askui, line 186: What do you think about dropping support for models that don’t support adaptive thinking and simplifying the logic?

mlikasam-askui (approved):

mlikasam-askui (approved):

src/askui/models/shared/thinking.py
philipph-askui: YOu are right, this is something we should do at some point. I think I would do that as soon as these models are no longer available (which is actually soon). Until then, I would keep this :-)

